PR #9, opened 2022-07-04
Add WatchResults endpoint

lemaitre-aneo:
This endpoint allows a user to watch a list of results and be notified when there is any change.
The user sends the list of ids they want to watch.
The submitter will then send the statuses for all requested ids immediately and keep the stream open.
The submitter will send updates to the client via the opened stream.
Any reply can be implicitely chunked if there are too many event to report at the same time (or for the first reply).
It is possible to filter out specific statuses from events with options


lemaitre-aneo:
Implements https://github.com/aneoconsulting/ArmoniK/issues/333

Protos/V1/submitter_service.proto
```diff
@@ -149,4 +159,42 @@ message WaitRequest {
...
+    string result_id = 1;
+    // Status of the result
+    ResultStatus.ResultStatus status = 2;
+  }
+
+  // List of result ids and their statuses
+  repeated ResultIdStatus statuses = 1;
+  // List of ids that were not found
+  // Required in stream mode
+  repeated string not_found_ids = 2;
```
wkirschenmann, line 199: Not fan of having this field defined in each message. Not sure whether a oneof would be a good idea here...
In case the message needs to be chuncked, the protocol could ensure that all not_found_ids are sent before the first ResultIdStatus...

wkirschenmann (changes requested):
WaitForAvailability should be removed.
It can very easily be implemented in API or in samples.

I am also wondering whether it would not be a good idea to have a stream for the watchResultrequest. It would allow either to chunk the list of result wanted or to add other result along with the submition of new tasks

Protos/V1/submitter_service.proto
```diff
@@ -149,4 +159,42 @@ message WaitRequest {
...
+ * Request to watch result states
+ * It contains the list of result ids you want to watch
+ *   and some options to filter out some events
+ */
+message WatchResultRequest {
+  message Options {
+    // List of statuses to ignore in the replies
+    repeated ResultStatus.ResultStatus ignore_statuses = 1;
+    // Skip the sending of result states at the beginning
+    optional bool skip_current_state = 2;
```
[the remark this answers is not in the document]
lemaitre-aneo: In theory, I agree with you, but I can imagine a few use cases where this would be useful. For instance when a user creates results, starts to watch, and only then starts the tasks that will produce those results.

Protos/V1/submitter_service.proto
lemaitre-aneo: According to our last conversation, an empty list is the same as the field not being here at all. So I don't really see the difference. Of course, in practice, the not_found_ids will only be sent at the beginning because it will not be retried.

lemaitre-aneo:
> WaitForAvailability should be removed.
> It can very easily be implemented in API or in samples.

Agreed.

> I am also wondering whether it would not be a good idea to have a stream for the watchResultrequest. It would allow either to chunk the list of result wanted or to add other result along with the submition of new tasks

Could be nice indeed. I will try something.

Protos/V1/submitter_service.proto
wkirschenmann: > Of course, in practice, the not_found_ids will only be sent at the beginning because it will not be retried.
Shouldn't this be enforced ?

lemaitre-aneo: Documented? definitely. Enforced? no.

Protos/V1/submitter_service.proto
```diff
@@ -149,4 +163,30 @@ message WaitRequest {
...
+ */
+message WatchResultRequest {
+  // result ids to fetch/watch
+  repeated string result_ids = 1;
+  // list of statuses to check results against
+  repeated ResultStatus.ResultStatus target_statuses = 2;
+  // Whether the current state should be fetched
+  bool fetch_current_state = 3;
+  // Whether the future state should be watched
+  bool watch_future_state = 4;
```
wkirschenmann, line 181: Why not use an enum instead of these two bools ?

Protos/V1/result_status.proto
```diff
@@ -11,4 +11,5 @@ enum ResultStatus
   RESULT_STATUS_CREATED = 1;
   RESULT_STATUS_COMPLETED = 2;
   RESULT_STATUS_ABORTED = 3;
+  RESULT_STATUS_NOTFOUND = 9;
```
aneojgurhem, line 14: Pourquoi 9 et pas 4 ?

Protos/V1/submitter_service.proto
```diff
@@ -23,11 +23,25 @@ service Submitter{
...
+	rpc WaitForAvailability (ResultRequest) returns (AvailabilityReply) {
+    option deprecated = true;
+  }
 	rpc WaitForCompletion (WaitRequest) returns (Count);
 	rpc CancelTasks (TaskFilter) returns (Empty); 
 	rpc GetTaskStatus (GetTaskStatusRequest) returns (GetTaskStatusReply);
-	rpc GetResultStatus (GetResultStatusRequest) returns (GetResultStatusReply);
+	rpc GetResultStatus (GetResultStatusRequest) returns (GetResultStatusReply) {
+    option deprecated = true;
+  }
```
aneojgurhem, line 34: cette indentation quoi
/s

Protos/V1/submitter_service.proto
```diff
@@ -149,4 +163,28 @@ message WaitRequest {
...
+  repeated string result_ids = 3;
+}
+
+/**
+ * List of Result statuses
+ */
+message WatchResultStream {
+  // Status of the results
+  ResultStatus.ResultStatus status = 1;
+  // List of result ids that triggered the event
```
aneojgurhem, line 188: L'idee c'est de retourner un message par statut voire plusieurs s'il y a trop d'ids d'un coup ? Je pense qu'on pourrait ajouter un peu de détails dans la doc de ce message. ça a été fait pour le message précédent.

aneojgurhem (changes requested):

Protos/V1/result_status.proto
lemaitre-aneo: Pour laisser de la place si on veux rajouter d'autres status. Idéalement, j'aurais mis `-1`, mais je ne sais pas si c'est possible.

aneojgurhem: ok parce que le notfound ne sera pas en base ?

lemaitre-aneo: c'est ça

aneojgurhem (approved):

Protos/V1/submitter_service.proto
wkirschenmann: Je ne parlerais pas de chunking du tout ! Il n'y a pas de différence entre envoyer 2x plein d'ids et 1x plein d'Id et juste après plein d'Id qui viennent de changer de statut.

lemaitre-aneo: maintenant explicité

Protos/V1/result_status.proto
lemaitre-aneo: remplacé par 127 pour avoir un encodage efficace

Protos/V1/submitter_service.proto
wkirschenmann: My proposal : "Result Ids are grouped by status. One message contains result Ids that have the same status."

Protos/V1/submitter_service.proto
lemaitre-aneo: explicited

PR #19, opened 2022-07-22
Package names should be in lowercase

jfonseca-aneo:


Protos/V1/result_status.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package ArmoniK.api.grpc.v1.ResultStatus;
+package armonik.api.grpc.v1.resultstatus;
```
lemaitre-aneo, line 3: `armonik.api.grpc.v1.result_status`

lemaitre-aneo (approved):

PR #21, opened 2022-07-25
Add partition names to Session and TaskOptions

lemaitre-aneo:


Protos/V1/objects.proto
```diff
@@ -15,6 +15,7 @@ message TaskOptions {
 	google.protobuf.Duration max_duration = 2;
 	int32 max_retries = 3;
 	int32 priority = 4;
+  string partition = 5;
```
jfonseca-aneo, line 18: partition_id?

wkirschenmann: I'd say `partition_id`

jfonseca-aneo: Ok; let's leave as partition

lemaitre-aneo: `partition_id` it is, then

jfonseca-aneo (approved):

PR #27, opened 2022-08-16
Remove submitter from worker and move session object

dbrasseur-aneo:
The worker api currently has access to the submitter service. Afaik, this shouldn't be the case, as it should instead use the agent service. If it's here for a good reason, I'll ammend the PR to only move the session object to objects.proto

### DIFF
--- a/Api/csharp/ArmoniK.Api.Worker/ArmoniK.Api.Worker.csproj
+++ b/Api/csharp/ArmoniK.Api.Worker/ArmoniK.Api.Worker.csproj
@@ -46,9 +46,6 @@
     <Protobuf Include="..\..\..\Protos\V1\agent_service.proto" GrpcServices="Both">
       <Link>gRPC\Protos\agent_service.proto</Link>
     </Protobuf>
-    <Protobuf Include="..\..\..\Protos\V1\submitter_service.proto" GrpcServices="Both">
-      <Link>gRPC\Protos\submitter_service.proto</Link>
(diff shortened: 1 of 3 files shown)


dbrasseur-aneo:
Should we also remove the monitoring related protos ? Like sessions_service ?

Api/csharp/ArmoniK.Api.Worker/ArmoniK.Api.Worker.csproj
```diff
@@ -46,9 +46,6 @@
     <Protobuf Include="..\..\..\Protos\V1\agent_service.proto" GrpcServices="Both">
       <Link>gRPC\Protos\agent_service.proto</Link>
     </Protobuf>
-    <Protobuf Include="..\..\..\Protos\V1\submitter_service.proto" GrpcServices="Both">
```
aneojgurhem, line 49: We need it. It is the packge used to implement this service in core. Unfortunately, it needs to have almost everything to avoid duplicate classes and just conflicts everywhere.

Protos/V1/objects.proto
```diff
@@ -18,6 +18,10 @@ message TaskOptions {
 	string partition_id = 5;
 }
 
+message Session {
+	string id = 1;
```
aneojgurhem, line 22: Why ? Does not seem necessary considering the previous comment.

aneojgurhem (changes requested):

PR #39, opened 2022-09-16
Improve submitter client ext

aneojgurhem:
fix https://github.com/aneoconsulting/ArmoniK/issues/512


Api/csharp/ArmoniK.Api.Client/Submitter/ReadOnlyByteArrayPayload.cs
```diff
@@ -0,0 +1,59 @@
...
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+using Google.Protobuf;
+
+namespace ArmoniK.Api.Client.Submitter
+{
+  public class ReadOnlyByteArrayPayload : IPayload
```
lemaitre-aneo, line 34: As far as I can tell, you can create a `ReadOnlyMemory<byte>` from a `byte[]`, so I would implement `ReadOnlyByteArrayPayload` in term of `ReadOnlyByteMemory`.

Api/csharp/ArmoniK.Api.Client/Submitter/StreamPayload.cs
```diff
@@ -0,0 +1,62 @@
...
+    public StreamPayload(Stream data)
+      => data_ = data;
+
+    public async IAsyncEnumerable<ByteString> ToChunkedByteStringAsync(int                                        maxChunkSize,
+                                                                       [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+      while (true)
+      {
+        var buffer = new byte[maxChunkSize];
+        var size = await data_.ReadAsync(buffer,
```
lemaitre-aneo, line 48: Wouldn't it be better to group multiple reads in case they are smaller than chunk size?

Api/csharp/ArmoniK.Api.Client/Submitter/ReadOnlyByteMemoryPayload.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+  public class ReadOnlyByteMemoryPayload : IPayload
+  {
+    private readonly ReadOnlyMemory<byte> readOnlyMemory_;
+
+    public ReadOnlyByteMemoryPayload(ReadOnlyMemory<byte> readOnlyMemory)
+      => readOnlyMemory_ = readOnlyMemory;
+
+#pragma warning disable CS1998
+    public async IAsyncEnumerable<ByteString> ToChunkedByteStringAsync(int maxChunkSize,
```
lemaitre-aneo, line 42: Maybe you can just create an `IEnumerable` and just let LINQ transform it into an `IAsyncEnumerable`

Api/csharp/ArmoniK.Api.Client/Submitter/SubmitterClientExt.cs
```diff
@@ -225,17 +180,52 @@ await stream.RequestStream.CompleteAsync()
...
+    /// <param name="client">gRPC client to the Submitter</param>
+    /// <param name="resultRequest">Request for result</param>
+    /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+    /// <returns>
+    ///   A stream in which the data will be written
+    /// </returns>
+    /// <exception cref="Exception">a result reply chunk is not data, rending it impossible to reconstitute the data</exception>
+    /// <exception cref="ArgumentOutOfRangeException">result reply type is unknown</exception>
+    [PublicAPI]
+    public static async Task<MemoryStream> GetResultAsStreamAsync(this gRPC.V1.Submitter.Submitter.SubmitterClient client,
```
lemaitre-aneo, line 197: I dont think it works as expected. From what I understand, you create and populate the stream entirely before returning it.

Api/csharp/ArmoniK.Api.Client/Submitter/StreamPayload.cs
aneojgurhem: It reads less than the buffer only if there is less bytes in the stream. I'm not sure that it is possible to have multiple smaller one

Api/csharp/ArmoniK.Api.Client/Submitter/ReadOnlyByteMemoryPayload.cs
aneojgurhem: You mean, create another method that returns `IEnumberable` and use `ToAsyncEnumerable` ?

lemaitre-aneo: Yes, as an implementation details at least. I don't know if it is of any value to add it to the interface though.

Api/csharp/ArmoniK.Api.Client/Submitter/StreamPayload.cs
lemaitre-aneo: Yes, but depending what is behind the stream (a socket for instance), you can have multiple small reads. I am not sure if it would be a problem in practice, though.

Api/csharp/ArmoniK.Api.Client/Submitter/ReadOnlyByteMemoryPayload.cs
aneojgurhem: Since it is internal, it does not really matter to add it to the interface

Api/csharp/ArmoniK.Api.Client/Internals/ReadOnlyByteMemoryPayload.cs
```diff
@@ -0,0 +1,65 @@
...
+    public ReadOnlyByteMemoryPayload(ReadOnlyMemory<byte> readOnlyMemory)
+      => readOnlyMemory_ = readOnlyMemory;
+
+    public IAsyncEnumerable<ByteString> ToChunkedByteStringAsync(int               maxChunkSize,
+                                                                 CancellationToken cancellationToken = default)
+      => ToChunkedByteString(maxChunkSize,
+                             cancellationToken)
+        .ToAsyncEnumerable();
+
+    private IEnumerable<ByteString> ToChunkedByteString(int               maxChunkSize,
```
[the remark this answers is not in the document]
aneojgurhem: Yeah, I forgot about it too... It is so useful.

Api/csharp/ArmoniK.Api.Client.Tests/SubmitterClientExtTest.cs
```diff
@@ -0,0 +1,122 @@
...
+  public void ChunkingWithCancel(IPayload payload,
+                                 byte[]   bytes,
+                                 int      maxChunkSize)
+  {
+    var res = new byte[bytes.Length];
+    var idx = 0;
+
+    var cancellationTokenSource = new CancellationTokenSource();
+
+    Assert.ThrowsAsync<OperationCanceledException>(async () =>
```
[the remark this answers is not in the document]
aneojgurhem: Cancellation does not really happen at the same time in the different implementations therefore you may not have the exact same number bytes written.

Api/csharp/ArmoniK.Api.Client/Internals/StreamPayload.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+      while (true)
+      {
+        var buffer = new byte[maxChunkSize];
+        var length = 0;
+
+        while (length < maxChunkSize)
+        {
+          cancellationToken.ThrowIfCancellationRequested();
+          var size = await data_.ReadAsync(buffer,
```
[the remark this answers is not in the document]
aneojgurhem: I thought this offset was used to write in the buffer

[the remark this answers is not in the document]
lemaitre-aneo: My bad, you are actually right.

Api/csharp/ArmoniK.Api.Client/Submitter/SubmitterClientExt.cs
```diff
@@ -225,39 +180,47 @@ await stream.RequestStream.CompleteAsync()
...
     /// <param name="resultRequest">Request for result</param>
     /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
     /// <returns>
-    ///   A byte array containing the data associated to the result
+    ///   A stream in which the data will be written
+    /// </returns>
+    /// <exception cref="Exception">a result reply chunk is not data, rending it impossible to reconstitute the data</exception>
+    /// <exception cref="ArgumentOutOfRangeException">result reply type is unknown</exception>
+    [PublicAPI]
+    public static Task<Stream> GetResultAsStreamAsync(this gRPC.V1.Submitter.Submitter.SubmitterClient client,
```
lemaitre-aneo, line 197: @aneojgurhem With the stream correctly implemented, it is not necessary for this function to be async. Do we remove the async?

wkirschenmann:
Not sure. Let's see how errors will be handled and so on before changing that.

Api/csharp/ArmoniK.Api.Client/Submitter/SubmitterClientExt.cs
aneojgurhem: We can keep it for now

Api/csharp/ArmoniK.Api.Client/Internals/AsyncEnumerableStream.cs
```diff
@@ -0,0 +1,204 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace ArmoniK.Api.Client.Internals
+{
+  /// <summary>
+  ///   IO.Stream that is read from an IAsyncEnumerable
+  /// </summary>
+  public class AsyncEnumerableStream : Stream, IAsyncDisposable
```
wkirschenmann, line 36: Lots of hypothesis on how it will be consumed.
Many of them will likely appear to be wrong.
This should be in another PR with lots of test and with microbenchmarks corresponding to different consumption use cases. These benchmarks should include both asynchronous and synchronous scenarios. In real application, the later are more likely as legacy deserializers are synchronous. None of the existing libs are using Memory or Span to get the data from the Stream.

wkirschenmann (changes requested):

aneojgurhem:
Not relevant anymore

PR #42, opened 2022-09-19
feat: create proto for applications

esoubiran-aneo:
All of these change are with the fact that the API can be used with the Admin.GUI and other client like the Admin.CLI.

Some explanations about proto and the response from the API.

`LastSessionsOptions last_sessions_options`: If not set, server must not return any sessions. If set, server will use `page` and `page_size` to return correct data (default to 0 for `page` and 3 for `page_size`). This is used to avoid each application to fetch server to know about last sessions.


Protos/V1/sessions_common.proto
```diff
@@ -8,8 +8,8 @@ import "session_status.proto";
 
 option csharp_namespace = "ArmoniK.Api.gRPC.V1.Sessions";
 
-/**
```
aneojgurhem, line 11: ça n'a pas lieu d'etre ici !

esoubiran-aneo:
Documentation is missing!

Protos/V1/sessions_common.proto
esoubiran-aneo: Effectivement, il y a un commit qui se balade et qui n'a rien à faire là

aneojgurhem: tu as manqué ce fichier aussi^^

PR #44, opened 2022-09-19
feat: create applications proto

esoubiran-aneo:
All of these change are with the fact that the API can be used with the Admin.GUI and other client like the Admin.CLI.

Some explanations about proto and the response from the API.

`LastSessionsOptions last_sessions_options`: If not set, server must not return any sessions. If set, server will use `page` and `page_size` to return correct data (default to 0 for `page` and 3 for `page_size`). This is used to avoid each application to fetch server to know about last sessions.

fix https://github.com/aneoconsulting/ArmoniK/issues/437


Protos/V1/applications_common.proto
```diff
@@ -0,0 +1,101 @@
...
+    ORDER_BY_FIELD_UNSPECIFIED = 0;
+    ORDER_BY_FIELD_ID = 1;
+    ORDER_BY_FIELD_TASKS_PENDING = 2;
+    ORDER_BY_FIELD_TASKS_IN_ERROR = 3;
+    ORDER_BY_FIELD_TASKS_COMPLETED = 4;
+    ORDER_BY_FIELD_TASKS_PROCESSING = 5;
+  }
+
+  // Options for the last sessions. If not specified, last sessions are not returned
+  LastSessionsOptions last_sessions_options = 3;
```
esoubiran-aneo, line 82: Could be in achieve in another request (using sessions endpoint with a limit)

Protos/V1/applications_common.proto
```diff
@@ -0,0 +1,101 @@
...
+import "google/protobuf/timestamp.proto";
+import "objects.proto";
+import "sessions_common.proto";
+
+option csharp_namespace = "Armonik.Api.Grpc.V1.Applications";
+
+/**
+ * An application object
+ */
+message Application {
```
esoubiran-aneo, line 14: Must use the new documentation convention

Protos/V1/applications_common.proto
```diff
@@ -0,0 +1,101 @@
...
+  int64 tasks_pending = 2;
+  // Number of tasks with a status "ERROR"
+  int64 tasks_in_error = 3;
+  // Number of tasks with a status "COMPLETED"
+  int64 tasks_completed = 4;
+  // Number of tasks with a status "PROCESSING"
+  int64 tasks_processing = 5;
+
+  // Last session related to the application
+  repeated SessionSummary session = 6;
```
esoubiran-aneo, line 38: Remove

Protos/V1/applications_common.proto
```diff
@@ -0,0 +1,101 @@
...
+   */
+  message ApplicationId {
+    // Application name
+    string name = 1;
+    // Application version
+    string version = 2;
+  }
+
+  // Number of tasks with a pending status ("CREATING", "SUBMITTED", "DISOATCHED")
+  int64 tasks_pending = 2;
```
esoubiran-aneo, line 29: Can be done in another request using a task endpoint

Protos/V1/applications_common.proto
esoubiran-aneo: done

PR #77, opened 2022-11-02
Add timeout and do not fail all builds when one fails

aneojgurhem:
Update pipeline for small fixes:
- Add timeout
- Do not fail all builds when one fails


lemaitre-aneo (approved):
/s Unbelievable, a short PR

PR #94, opened 2023-01-09
ci: update workflows

### LINKED ISSUE #93: Update workflows files
use this https://github.com/aneoconsulting/ArmoniK/discussions/822

esoubiran-aneo:
This will allow CI to fail faster if an error occured and rename some part to improve readibility

fix #93

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,6 @@
 name: CI
 
 on:
-  push:
-    branches:
-      - main
   pull_request:
     branches:
       - main
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,31 @@
 name: CI
 
 on:
+  push:
+    branches:
+      - main
   pull_request:
+    branches:
+      - main
 
 jobs:
+  format-protobuf:
+    name: Format Protobuf
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.ref }}
+          fetch-depth: 0
+          submodules: true
+
+      - name: Check Diff
+        run: docker run --rm --volume "$(pwd)/Protos/V1:/workspace" --workdir /workspace bufbuild/buf:1.8.0 format -d --exit-code
+
   format-csharp:
+    needs: format-protobuf
     name: Format C#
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -2,9 +2,27 @@ name: CI
...
+        with:
+          ref: ${{ github.ref }}
+          fetch-depth: 0
+          submodules: true
+
+      - name: Check Diff
+        run: docker run --rm --volume "$(pwd)/Protos/V1:/workspace" --workdir /workspace bufbuild/buf:1.8.0 format -d --exit-code
+
   format-csharp:
+    needs: format-protobuf
```
aneojgurhem, line 25: This one takes so much time that I prefer having failed rather than waiting for it

.github/workflows/release-packages.yml
```diff
@@ -1,37 +1,49 @@
...
-      - name: Generate Version
-        id: genver
+      - name: Generate version from git
+        if: ${{ !contains(github.ref, 'refs/tags') }}
+        run: |
+          VERSION=$(docker run --rm -v $(pwd):/repo codacy/git-version /bin/git-version --folder=/repo --release-branch=main)
+          echo "version=$VERSION" >> $GITHUB_ENV
+
+      - name: Get version from tag
+        if: ${{ contains(github.ref, 'refs/tags') }}
```
aneojgurhem, line 30: How do you ensure that the tag is coherent ?
This should work only on X.Y.Z tags. Moreover, how do you ensure that the tag given is really the next tag ?
Things I do not want to see:
0.6.6 -> 0.8.0
0.6.6 -> 0.6.8

esoubiran-aneo: I can't. Maintainers are responsible for publishing the correct tag (and you can do this using a PR with a review). Using an automated system is the same, you have the responsibility to add `feature` or `breaking` in a comment from one or more commit. So, if another maintainer forgot to add `breaking` in the commit message, the tag will not be correct and the workflow can be achieved by only one person. With a tag release from a PR, you must have a pair validation which I think is better.

.github/workflows/ci.yml
esoubiran-aneo: Need to be changed

aneojgurhem:
As I do not understand completely how to contribute and make a release, maybe it is time to fill in the HOW-TO-CONTRINUTE ?

PR #195, opened 2023-03-16
feat: add health checks

### LINKED ISSUE #168: Add health check api
In order to know from the GUI if control_plane is still up, we need an health check. We can use API or the control_plane health check using http from the ingress. Waiting to know the better solution in #195

esoubiran-aneo:
fix #168

This will allow connected client to know is the API is up and running (so client can communicate with it) and if the cluter is up and running.

### DIFF
--- a/Protos/V1/health_checks_common.proto
+++ b/Protos/V1/health_checks_common.proto
@@ -7,7 +7,7 @@ option csharp_namespace = "Armonik.Api.Grpc.V1.HealthChecks";
 /**
  * Represents the available health status
  */
- enum HealthStatusEnum {
+enum HealthStatusEnum {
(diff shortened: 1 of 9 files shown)


esoubiran-aneo:
Could be achieved using ingress @lemaitre-aneo (using the core http health check) ?

lemaitre-aneo:
This could be done via ingress only to have the health of the particuliar control plane pod your accessing. I think it is much better to have the health check embedded in the API.

esoubiran-aneo:
> This could be done via ingress only to have the health of the particuliar control plane pod your accessing. I think it is much better to have the health check embedded in the API.

Yes, I agree!

lemaitre-aneo (dismissed):

lemaitre-aneo (commented):
You should add it to the mock

lemaitre-aneo (approved):

PR #198, opened 2023-03-21
feat: add missing methods for partitions management

aneojgurhem:
Add the missing methods for partition management in the partitions service:
- add partition
- remove partition
- update partition

### DIFF
--- a/Protos/V1/partitions_common.proto
+++ b/Protos/V1/partitions_common.proto
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 package armonik.api.grpc.v1.partitions;
 
+import "objects.proto";
+
 option csharp_namespace = "Armonik.Api.Grpc.V1.Partitions";
 
 /**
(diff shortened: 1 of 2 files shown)


Protos/V1/partitions_common.proto
```diff
@@ -115,3 +117,82 @@ message GetPartitionRequest {
...
+ * Request to update multiple partitions.
+ *
+ * Cannot update more than 30 partitions
+ */
+message UpdatePartitionsRequest {
+  /**
+   * Represents the data of the partitions that will be updated.
+   *
+   * ID is mandatory.
+   * -1 means that value will not be updated
```
lemaitre-aneo, line 131: I would just say negative instead of specifically -1.
The problem is that in protobuf, if a value is missing, it is set to 0...

Protos/V1/partitions_common.proto
```diff
@@ -115,3 +117,82 @@ message GetPartitionRequest {
...
+ */
+message CreatePartitionsRequest {
+  /**
+   * Represents the data of the partitions that will be created.
+   */
+  message PartitionCreate {
+    repeated string parent_partition_ids = 2; /** The parent partition IDs. */
+    int64 pod_reserved = 3; /** Whether the partition is reserved for pods. */
+    int64 pod_max = 4; /** The maximum number of pods that can be used by sessions using the partition. */
+    map<string, string> pod_configuration = 7; /** The pod configuration. */
```
lemaitre-aneo, line 183: As far as I remember, `map` is not well supported. I think we used a list of pairs in other places.

Protos/V1/partitions_common.proto
aneojgurhem: Ok.
But 0 is a valid number for our use case...

esoubiran-aneo (changes requested):
Waiting for implentation

PR #199, opened 2023-03-21
ci: stop two workflows from cancelling each others

aneojgurhem:
Stop CI and Validate Doc workflows from cancelling each other since they have the same concurrency group.


esoubiran-aneo (approved):
LGTM 👍

PR #204, opened 2023-03-30
feat: Add alternative methods to create tasks and results separately

aneojgurhem:
Adds rpc methods in TasksService, ResultsService and AgentService for task and result dissociation.


Protos/V1/agent_common.proto
```diff
@@ -70,13 +72,145 @@ message Result {
...
+  string communication_token = 3; /** Communication token received by the worker during task processing */
+}
+
+/**
+* Result metadata
+*/
+message ResultMetaData {
+  string session_id = 1; /** The session ID. */
+  string name = 2; /** The result name. */
+  string owner_task_id = 3; /** The owner task ID. */
```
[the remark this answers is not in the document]
aneojgurhem: It will be empty after result metada creation

Protos/V1/agent_common.proto
```diff
@@ -70,13 +72,145 @@ message Result {
...
+  * - id
+  * - data_chunk (stream can have multiple data_chunk messages that represent data divided in several parts)
+  * - data_complete
+  *
+  * Data chunk cannot exceed the size returned by the GetServiceConfiguration rpc method
+  */
+  oneof type {
+    ResultIdentifier id = 1; /** The identifier of the result to which add data. */
+    bytes data_chunk = 2; /** A chunk of data. */
+    bool data_complete = 3; /** Last message telling the data are complete. */
```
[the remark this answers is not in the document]
aneojgurhem: So it will be different from the other streams we designed ?

Protos/V1/agent_common.proto
```diff
@@ -70,13 +72,145 @@ message Result {
...
+*/
+message CreateResultsMetaDataRequest {
+  /**
+  * A result to create.
+  */
+  message ResultCreate {
+    string name = 1; /** The result name. Given by the client. */
+  }
+  repeated ResultCreate results = 1; /** The list of results to create. */
+  string session_id = 2; /** The session in which create results. */
```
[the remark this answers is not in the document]
aneojgurhem: We needed this information when the client was able to create the result_id. I think we can have them for consistency.

Protos/V1/agent_common.proto
```diff
@@ -16,14 +18,21 @@ message CreateTaskRequest {
...
 
-  string communication_token = 4;
+  string communication_token = 4; /** Communication token received by the worker during task processing */
 }
 
 message CreateTaskReply {
+  message ResultIdName {
+    string name = 1; /** Result given name. */
+    string result_id = 2; /** Result Unique ID. */
+  }
```
dbrasseur-aneo, line 28: Why not use a map ? https://protobuf.dev/programming-guides/encoding/#maps

Protos/V1/agent_common.proto
```diff
@@ -70,13 +79,152 @@ message Result {
...
+* expected_output_ids and data_dependencies must be created through ResultsService.
+*
+* Remark : this may have to be enriched to a better management of errors but
+* will the client application be able to manage a missing data dependency or expected output ?
+*/
+message SubmitTasksResponse {
+  message ResultIdName {
+    string name = 1; /** Result given name. */
+    string result_id = 2; /** Result Unique ID. */
+  }
```
dbrasseur-aneo, line 156: Why not use a map ? https://protobuf.dev/programming-guides/encoding/#maps

Protos/V1/agent_common.proto
```diff
@@ -70,13 +79,152 @@ message Result {
...
+
+  message TaskInfo {
+    string task_id = 1; /** The task ID. */
+    repeated string expected_output_ids = 2; /** The expected output IDs. A task have expected output IDs. */
+    repeated string data_dependencies = 3; /** The data dependencies IDs (inputs). A task have data dependencies. */
+    string payload_id = 4; /** Unique ID of the result that will be used as payload. Results are created implicitly. */
+    repeated ResultIdName id_name = 5; /** List of association between given names and result IDs. */
+  }
```
dbrasseur-aneo, line 164: Why have multiple definitions of taskinfo in different message types ?

Protos/V1/results_common.proto
```diff
@@ -104,18 +113,139 @@ message ListResultsResponse {
...
+
+/*
+* Request for creating results without data
+*/
+message CreateResultsMetaDataRequest {
+  /**
+  * A result to create.
+  */
+  message ResultCreate {
+    string name = 1; /** The result name. Given by the client. */
```
dbrasseur-aneo, line 140: Why have a type with only 1 field ? Futureproofing ?

Protos/V1/submitter_common.proto
```diff
@@ -52,10 +52,17 @@ message CreateLargeTaskRequest {
 }
 
 message CreateTaskReply {
+  message ResultIdName {
+    string name = 1; /** Result given name. */
+    string result_id = 2; /** Result Unique ID. */
+  }
```
dbrasseur-aneo, line 58: Why not use a map ?

Protos/V1/submitter_common.proto
```diff
@@ -52,10 +52,17 @@ message CreateLargeTaskRequest {
...
+
   message TaskInfo {
-    string task_id = 1;
-    repeated string expected_output_keys = 2;
-    repeated string data_dependencies = 3;
+    string task_id = 1; /** Unique ID of the created task. */
+    repeated string expected_output_keys = 2; /** Unique ID of the result that will be used as expected output. Results are created implicitly. */
+    repeated string data_dependencies = 3; /** Unique ID of the result that will be used as data dependency. Results should already exist. */
+    string payload_id = 4; /** Unique ID of the result that will be used as payload. Results are created implicitly. */
+    repeated ResultIdName id_name = 5; /** List of association between given names and result IDs. */
```
dbrasseur-aneo, line 65: Same as before, couldn't it be in objects common ?

Protos/V1/tasks_common.proto
```diff
@@ -221,3 +235,37 @@ message CountTasksByStatusRequest {}
...
+/**
+* Request to create tasks.
+*/
+message SubmitTasksRequest {
+  message TaskCreation {
+    repeated string expected_output_keys = 1; /** Unique ID of the results that will be produced by the task. Results are created using ResultsService. */
+    repeated string data_dependencies = 2; /** Unique ID of the results that will be used as datadependencies. Results are created using ResultsService. */
+    string payload_id = 3; /** Unique ID of the result that will be used as payload. Results are created using ResultsService. */
+    TaskOptions task_options = 4; /** Optionnal task options. */
+  }
```
dbrasseur-aneo, line 248: Missing the id-name mappings

Protos/V1/agent_common.proto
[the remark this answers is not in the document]
lemaitre-aneo: As it seems to be used only upon result creation, this field will be necessarily empty in the reply, so there is no need to provide it.

Protos/V1/agent_common.proto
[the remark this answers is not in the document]
lemaitre-aneo: The question is: What do we want to keep in the future? I would say the simpler, the better. For the download, it become so simple that the oneof disappear completely.

Protos/V1/objects.proto
```diff
@@ -41,9 +41,10 @@ message Output {
 }
 
 message TaskRequest {
-  repeated string expected_output_keys = 1;
-  repeated string data_dependencies = 2;
-  bytes payload = 3;
+  repeated string expected_output_keys = 1; /** Given names to the expected outputs that will be created implicitly. IDs are returned after task creation */
+  repeated string data_dependencies = 2; /** IDs of the results that will be used as data dependency. */
+  bytes payload = 3; /** Content of the payload for the task. */
+  string payload_name = 4; /** Name that will be associated to the result created for the payload. */
```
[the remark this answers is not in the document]
lemaitre-aneo: Fair enough

Protos/V1/results_common.proto
lemaitre-aneo: Yes, this is common practice that rpc have only custom types

Protos/V1/agent_common.proto
```diff
@@ -70,13 +79,152 @@ message Result {
...
+  string session_id = 2; /** The session in which create results. */
+  string communication_token = 3; /** Communication token received by the worker during task processing */
+}
+
+/**
+* Result metadata
+*/
+message ResultMetaData {
+  string session_id = 1; /** The session ID. */
+  string result_id = 2; /** The session ID. */
```
dbrasseur-aneo, line 113:
```suggestion
  string result_id = 2; /** The result ID. */
```

Protos/V1/agent_common.proto
aneojgurhem: finally, we don't need this message

Protos/V1/agent_common.proto
```diff
@@ -16,14 +18,21 @@ message CreateTaskRequest {
...
+
   message TaskInfo {
-    string task_id = 1;
-    repeated string expected_output_keys = 2;
-    repeated string data_dependencies = 3;
+    string task_id = 1; /** The task ID. */
+    repeated string expected_output_keys = 2; /** The expected output IDs. A task have expected output IDs. */
+    repeated string data_dependencies = 3; /** The data dependencies IDs (inputs). A task have data dependencies. */
+    string payload_id = 4; /** Unique ID of the result that will be used as payload. Results are created implicitly. */
+    repeated ResultIdName id_name = 5; /** List of association between given names and result IDs. */
```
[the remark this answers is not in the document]
aneojgurhem: not needed as we expect results to be already existing

Protos/V1/agent_common.proto
aneojgurhem: mutualizing the message means introducing a breaking change in the existing message.
I'm not sure we want to introduce more coupling between messages.

Protos/V1/agent_common.proto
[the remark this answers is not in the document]
aneojgurhem: I'll remove it

Protos/V1/agent_common.proto
dbrasseur-aneo: Fair enough

Protos/V1/tasks_common.proto
aneojgurhem: not required anymore

Protos/V1/tasks_common.proto
```diff
@@ -221,3 +235,36 @@ message CountTasksByStatusRequest {}
...
+
+/**
+* Request to create tasks.
+*/
+message SubmitTasksRequest {
+  message TaskCreation {
+    repeated string expected_output_keys = 1; /** Unique ID of the results that will be produced by the task. Results are created using ResultsService. */
+    repeated string data_dependencies = 2; /** Unique ID of the results that will be used as datadependencies. Results are created using ResultsService. */
+    string payload_id = 3; /** Unique ID of the result that will be used as payload. Results are created using ResultsService. */
+  }
```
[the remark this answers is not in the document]
aneojgurhem: add

Protos/V1/agent_common.proto
aneojgurhem: @lemaitre-aneo your thougts ?

lemaitre-aneo: @aneojgurhem I agree with you. We should delay this discussion until we break compatibility.

Protos/V1/agent_common.proto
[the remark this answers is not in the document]
lemaitre-aneo: I'm not convinced about the consistency as I think the communication_token can replace the session id. In fact, it might be possible to factorize the client proto with the agent proto if we say that you have a session id for the client and a communication token for the agent.

But all in all, I don't really care.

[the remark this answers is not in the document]
aneojgurhem: We need to take that into account for the V4

Protos/V1/agent_common.proto
```diff
@@ -70,13 +73,144 @@ message Result {
...
+
+/**
+* Result metadata
+*/
+message ResultMetaData {
+  string session_id = 1; /** The session ID. */
+  string result_id = 2; /** The result ID. */
+  string name = 3; /** The result name. */
+  result_status.ResultStatus status = 4; /** The result status. */
+  google.protobuf.Timestamp created_at = 5; /** The result creation date. */
```
lemaitre-aneo, line 110: I think we could need an end time to have the time when the result has been completed or turn into an error. Could be a later addition.

Protos/V1/agent_common.proto
```diff
@@ -70,13 +73,144 @@ message Result {
...
+  /**
+  * The possible messages that constitute a UploadResultDataRequest
+  * They should be sent in the following order:
+  * - id
+  * - data_chunk (stream can have multiple data_chunk messages that represent data divided in several parts)
+  *
+  * Data chunk cannot exceed the size returned by the GetServiceConfiguration rpc method
+  */
+  oneof type {
+    ResultIdentifier id = 1; /** The identifier of the result to which add data. */
```
lemaitre-aneo, line 204: Why can't you just have result_id?

Protos/V1/agent_common.proto
```diff
@@ -70,13 +73,144 @@ message Result {
...
+message CreateResultsResponse {
+  repeated ResultMetaData results = 1; /** The raw results that were created. */
+  string communication_token = 2; /** Communication token received by the worker during task processing */
+}
+
+/*
+* Request for uploading results data through stream.
+* Data must be sent in multiple chunks
+*/
+message UploadResultDataRequest {
```
lemaitre-aneo, line 186: With such a stream message, it would be possible to support multiple uploads in the same stream without any addition to the message. Should we allow it, or forbid it? In any case, we should be explicit about it.

Protos/V1/results_common.proto
```diff
@@ -104,18 +113,137 @@ message ListResultsResponse {
...
+/*
+* Request for creating results without data
+*/
+message CreateResultsRequest {
+  /**
+  * A result to create.
+  */
+  message ResultCreate {
+    bytes data = 1; /** The actual data of the result. */
+    string name = 2; /** The result name. Given by the client. */
```
lemaitre-aneo, line 162: I would invert those to have `name = 1`, like in `CreateResultsMetaDataRequest`

Protos/V1/results_service.proto
```diff
@@ -19,4 +20,35 @@ service Results {
...
+
+  /**
+   * Create one result with data included in the request
+   */
+  rpc CreateResults(CreateResultsRequest) returns (CreateResultsResponse) {}
+
+  /**
+   * Upload data for result with stream
+   */
+  rpc UploadResultData(stream UploadResultDataRequest) returns (UploadResultDataResponse) {}
```
lemaitre-aneo, line 38: I thought you wanted to have an alternative call without stream for small results. Is it still something we would want?

packages/csharp/ArmoniK.Api.Client/ResultsClientExt.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// <param name="resultId">The id of the result</param>
+    /// <param name="data">The data to send</param>
+    /// <returns>
+    ///   The upload result response
+    /// </returns>
+    [PublicAPI]
+    public static async Task<UploadResultDataResponse> UploadResultData(this Results.ResultsClient client,
+                                                                        string                     sessionId,
+                                                                        string                     resultId,
+                                                                        byte[]                     data)
```
lemaitre-aneo, line 58: I'm not sure if we should take a `byte[]` or a `ReadOnlyMemory`

packages/csharp/ArmoniK.Api.Client/ResultsClientExt.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// </summary>
+    /// <param name="client">gRPC result client</param>
+    /// <param name="sessionId">The id of the session</param>
+    /// <param name="resultId">The id of the result</param>
+    /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+    /// <returns>
+    ///   A byte array containing the data associated to the result
+    /// </returns>
+    [PublicAPI]
+    public static async Task<byte[]> DownloadResultData(this Results.ResultsClient client,
```
lemaitre-aneo, line 108: Same here. Returning a `byte[]` or a `Memory`

Protos/V1/agent_common.proto
aneojgurhem: Let's create some issues before resolving this comment

Protos/V1/agent_common.proto
```diff
@@ -70,13 +73,144 @@ message Result {
...
+*
+* Remark : this may have to be enriched to a better management of errors but
+* will the client application be able to manage a missing data dependency or expected output ?
+*/
+message SubmitTasksResponse {
+  message TaskInfo {
+    string task_id = 1; /** The task ID. */
+    repeated string expected_output_ids = 2; /** The expected output IDs. A task has expected output IDs. */
+    repeated string data_dependencies = 3; /** The data dependencies IDs (inputs). A task has data dependencies. */
+    string payload_id = 4; /** Unique ID of the result that will be used as payload. Results are created implicitly. */
```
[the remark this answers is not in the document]
aneojgurhem: It is the principal way of finding which ID is generated from what was submitted.
Task status would be creating all the time at this point. Not that interesting IMO.

Protos/V1/agent_common.proto
aneojgurhem: SessionId is included in this field

Protos/V1/agent_common.proto
aneojgurhem: As the code is implemented, we support only one send

packages/csharp/ArmoniK.Api.Client/ResultsClientExt.cs
aneojgurhem: Still linked to #39

packages/csharp/ArmoniK.Api.Client/ResultsClientExt.cs
aneojgurhem: an effort for #39

lemaitre-aneo (approved):

PR #209, opened 2023-04-03
feat: add retry to task status

### LINKED ISSUE #206: Add a retry status on tasks

esoubiran-aneo:
fix #206


Protos/V1/task_status.proto
```diff
@@ -22,4 +22,5 @@ enum TaskStatus {
   TASK_STATUS_CANCELLED = 8; /** Task is cancelled. */
   TASK_STATUS_PROCESSING = 9; /** Task is being processed. */
   TASK_STATUS_PROCESSED = 10; /** Task is processed. */
+  TASK_STATUS_RETRIED = 11; /** Task is retried. */
```
lemaitre-aneo, line 25: This is not a real status for now. What would be the use for such a status? What semantics to we want for it?

aneojgurhem: This status would help distinguish tasks that are in error from too much retries (error) from task in error but retried and completed.
So retried tasks will be in retried status and last try would either be completed or error.

lemaitre-aneo (approved):

PR #211, opened 2023-04-03
feat: create meta

### LINKED ISSUE #207: Add a generic way to filter
Multiples issues are related to the need of generics filters.

In fact, we want the user to be able to apply filter on any column. Some column will be known but options from the tasksOptions is just a map<string, string>
Also, how do I can get the schema of an item?

esoubiran-aneo:
fix #207

### DIFF
--- a/Protos/V1/meta.proto
+++ b/Protos/V1/meta.proto
@@ -1,0 +1,17 @@
+syntax = "proto3";
+
+package armonik.api.grpc.v1.meta;
+
+import "order_direction.proto";
+
+option csharp_namespace = "Armonik.Api.Grpc.V1.Meta";
+
+message Meta {
+  int32 page = 1; /** The page number. Start at 0.*/
+  int32 page_size = 2; /** The page size. */
+
+  string order_by = 3; /** The order by. */
+  order_direction.OrderDirection order = 4; /** The order. */
+
+  map<string, string> filters = 5; /** The filters. */
+}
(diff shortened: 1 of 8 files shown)


lemaitre-aneo (commented):
This is a nice change in theory, but it is a real breaking change. I am not sure it is worth it for now. Maybe keep the idea when we will break the API for real.

Also, I'm not convinced about the name Meta. If we want to keep a generic name, I would prefer common or utils, but in your case, I would go with Pager or something like that.

Protos/V1/meta.proto
```diff
@@ -0,0 +1,17 @@
...
+option csharp_namespace = "Armonik.Api.Grpc.V1.Meta";
+
+message Meta {
+  int32 page = 1; /** The page number. Start at 0.*/
+  int32 page_size = 2; /** The page size. */
+
+  string order_by = 3; /** The order by. */
+  order_direction.OrderDirection order = 4; /** The order. */
+
+  map<string, string> filters = 5; /** The filters. */
```
Barbapapazes, line 16: Need to add more detail about the usage (for exemple for the usage of the end and start date)

esoubiran-aneo:
> This is a nice change in theory, but it is a real breaking change. I am not sure it is worth it for now. Maybe keep the idea when we will break the API for real.
> 
> Also, I'm not convinced about the name Meta. If we want to keep a generic name, I would prefer common or utils, but in your case, I would go with Pager or something like that.

I will open a new PR in order to only fix #207 without other change.

esoubiran-aneo:
Will be implemented in other PR

PR #222, opened 2023-04-06
feat: Added mTLS support to C# Api

dbrasseur-aneo:
mTLS support was implemented incorrectly, this PR aims at correcting this problem and have a functional implementation.
Note that it *should* only support TLS 1.2 at most, as TLS 1.3 is not supported on netstandard.

### DIFF
--- a/packages/csharp/ArmoniK.Api.Common.Channel/ArmoniK.Api.Common.Channel.csproj
+++ b/packages/csharp/ArmoniK.Api.Common.Channel/ArmoniK.Api.Common.Channel.csproj
@@ -5,6 +5,8 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <PackageVersion>3.7.2</PackageVersion>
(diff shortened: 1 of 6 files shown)


packages/csharp/ArmoniK.Api.Common.Channel/ArmoniK.Api.Common.Channel.csproj
```diff
@@ -5,6 +5,8 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <PackageVersion>3.7.2</PackageVersion>
+    <PackageOutputPath>../publish</PackageOutputPath>
```
aneojgurhem, line 8: why do you need that ?

dbrasseur-aneo: Only some packages were published locally on build. That enables debugging with local publishing.

aneojgurhem (approved):

PR #226, opened 2023-04-12
fix:correct path for html python coverage

ngruelaneo:


.github/workflows/ci.yml
```diff
@@ -182,8 +182,8 @@ jobs:
       - name: Archive code coverage results html
         uses: actions/upload-artifact@v3
         with:
-          name: code-coverage-report-html
-          path: coverage_report
+          name: packages/python/code-coverage-report-html
```
dbrasseur-aneo, line 185:
```suggestion
          name: code-coverage-report-html
```

dbrasseur-aneo (changes requested):

.github/workflows/ci.yml
ngruelaneo: done!

dbrasseur-aneo (approved):

PR #227, opened 2023-04-18
fix: Fixed mTLS support for concurrent channels and .net framework support

dbrasseur-aneo:


packages/csharp/ArmoniK.Api.Client/Options/GrpcClient.cs
```diff
@@ -59,5 +65,23 @@ public class GrpcClient
...
+    ///   Path to the Certificate Authority file in pem format
+    /// </summary>
+    public string CaCert { get; set; } = "";
+
+    /// <summary>
+    ///   Override the endpoint name during SSL verification. This option is only used when AllowUnsafeConnection is true and
+    ///   only when the runtime is .NET Framework.
+    ///   Automatic target name by default. Should be overriden by the right name to reduce performance cost.
+    /// </summary>
+    public string OverrideTargetName { get; set; } = OverrideTargetNameAutomatic;
```
lemaitre-aneo, line 79: Couldn't you use the empty string to mean automatic?

packages/csharp/ArmoniK.Api.Client/Submitter/GrpcChannelFactory.cs
```diff
@@ -52,6 +55,150 @@ namespace ArmoniK.Api.Client.Submitter
...
+      return builder.ToString();
+    }
+
+    /// <summary>
+    ///   Exports a certificate to pem format
+    /// </summary>
+    /// <param name="cert">Certificate to export</param>
+    /// <returns>Certificate in pem formatted string</returns>
+    private static string ExportToPem(X509Certificate2 cert)
+      => $"-----BEGIN CERTIFICATE-----\n{Convert.ToBase64String(cert.GetRawCertData(), Base64FormattingOptions.InsertLineBreaks)}\n-----END CERTIFICATE-----";
```
lemaitre-aneo, line 84: Is there no such method on a certificate object already?

dbrasseur-aneo: In net standard, afaik, no. https://stackoverflow.com/a/43941142

lemaitre-aneo (approved):

PR #228, opened 2023-04-21
feat: add generics filters

### LINKED ISSUE #207: Add a generic way to filter

esoubiran-aneo:
fix #207

### DIFF
--- a/Protos/V1/applications_common.proto
+++ b/Protos/V1/applications_common.proto
@@ -32,7 +32,6 @@ message ListApplicationsRequest {
   int32 page = 1; /** The page number. Start at 0. */
   int32 page_size = 2; /** Number of items per page. */
 
-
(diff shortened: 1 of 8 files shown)


aneojgurhem (commented):
That's a big breaking change! How do you express filtering after a date ? It was possible to do so in the previous version. It does not seem to be that simple now.

Moreover, How do we use the new version ? What should I put in the map ? Should I know the database schema beforehand ? What if the schema changes ? What if we stop using mongodb ?

I would suggest to keep the existing filters (eventually add some more depending on needs) and add a specific map to filter on task options. It allows to decouple the Api and the database schema.

PR #230, opened 2023-04-24
ci: Added version check for release

dbrasseur-aneo:
When we do a release, this will check if we forgot to change the version of the packages, preventing issues where Python would send a new version to pypi but C# packages would be the old version

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,8 @@ on:
       - "[0-9]+.[0-9]+.[0-9]+"
 
 jobs:
+  # We need to verify the version in some packages with the tag in order to have the same version in all packages (python, csharp, angular, web).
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -9,7 +9,33 @@ on:
       - "[0-9]+.[0-9]+.[0-9]+"
 
 jobs:
+  # We need to verify the version in some packages with the tag in order to have the same version in all packages (python, csharp, angular, web).
```
esoubiran-aneo, line 12: @dbrasseur-aneo can you check this comment? If it's good, it will be merged.

esoubiran-aneo (changes requested):

.github/workflows/release.yml
dbrasseur-aneo: Looks good to me

PR #234, opened 2023-04-25
chore: remove unused vscode extension

esoubiran-aneo:
There is no yml file (others than Actions) so this extension is not necessary. If we want, we can add the GitHub Actions extension.


aneojgurhem:
This extension supports github workflow I think

esoubiran-aneo:
Yes, will add it.

PR #236, opened 2023-04-25
feat: add a service to return components versions

### LINKED ISSUE #223: Create an endpoint to get current version of each module
> Mettre la possibilité en configuration d’afficher à côté de la version de la GUI la version actuelle d’ArmoniK

esoubiran-aneo:
fix #223

### DIFF
--- a/Protos/V1/versions_common.proto
+++ b/Protos/V1/versions_common.proto
@@ -1,0 +1,22 @@
+/**
+ * Message describing versions.
+ */
+syntax = "proto3";
+
+package armonik.api.grpc.v1.versions;
+
+option csharp_namespace = "Armonik.Api.Grpc.V1.Versions";
+
+/**
+ * Request to list versions.
+ */
+message ListVersionsRequest {}
+
+/**
+ * Response to list versions.
+ */
+message ListVersionsResponse {
+    string armonik = 1;
+    string core = 2;
+    // We can add more versions here.
+}
--- a/Protos/V1/versions_service.proto
+++ b/Protos/V1/versions_service.proto
@@ -1,0 +1,21 @@
+/**
+ * Versions related methods within a service.
+ * This service will be used to get the version of infrastructure components from outside the cluster.
+ */
+syntax = "proto3";
+
+package armonik.api.grpc.v1.versions;
+
+import "versions_common.proto";
+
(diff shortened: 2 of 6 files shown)


aneojgurhem:
Not convinced. How do we implement it ?

dbrasseur-aneo:
I'd say a kubernetes command to get the image versions, but it doesn't belong to the api imo

aneojgurhem:
I think that this output should not be a responsibility of Core.

aneojgurhem (commented):
I think that this output should not be a responsibility of Core.

esoubiran-aneo:
> I think that this output should not be a responsibility of Core.

Maybe aqi should be out of core?

aneojgurhem:
A part of it should most likely be out of core

Protos/V1/versions_common.proto
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Request to list versions.
+ */
+message ListVersionsRequest {}
+
+/**
+ * Response to list versions.
+ */
+message ListVersionsResponse {
+  string armonik = 1;
```
esoubiran-aneo, line 19: Peut être passé via une var en via l'infra. (mais ça n'a pas forcément de sens parce que l'infra va être customisée).

Mais c'est aussi dans les logs de seq maintenant.
```suggestion
```

Protos/V1/versions_common.proto
```diff
@@ -0,0 +1,22 @@
...
+  string core = 2;
```
esoubiran-aneo, line 20:
```suggestion
  string core = 1;
```

Protos/V1/versions_common.proto
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Request to list versions.
+ */
+message ListVersionsRequest {}
+
+/**
+ * Response to list versions.
+ */
+message ListVersionsResponse {
+  string core = 1;
```
esoubiran-aneo, line 19:
```suggestion
  string core = 1;
  string api = 2;
```

aneojgurhem (approved):

esoubiran-aneo:
Could be done by using the reverse proxy which send a json (https://github.com/aneoconsulting/ArmoniK/blob/main/versions.tfvars.json).

We need to write documentation. If a 404 is returned, gui can show a link to documentation

PR #258, opened 2023-05-21
fix: retry and keep alive

ddubuc:
Need to re-add features to manage retry and keep alive in the tcp connection

### DIFF
--- a/packages/csharp/ArmoniK.Api.Client/Options/GrpcClient.cs
+++ b/packages/csharp/ArmoniK.Api.Client/Options/GrpcClient.cs
@@ -102,6 +102,12 @@ namespace ArmoniK.Api.Client.Options
     public int MaxAttempts { get; set; } = 5;
 
     /// <summary>
+    ///   The backoff will be multiplied by this multiplier after each retry attempt and will increase exponentially when the
+    ///   multiplier is greater than 1.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


packages/csharp/ArmoniK.Api.Client/Options/GrpcClient.cs
```diff
@@ -72,10 +72,41 @@ public class GrpcClient
...
+
+    /// <summary>
+    ///   KeepAliveTimeInterval is the interval in milliseconds at which the connection will be kept alive.
+    /// </summary>
+    public int KeepAliveTimeInterval { get; set; } = 30000;
+
+    /// <summary>
+    ///   MaxIdleTime is the maximum idle time in minutes after which the connection will be closed.
+    /// </summary>
+    public int MaxIdleTime { get; set; } = 5;
```
lemaitre-aneo, line 95: You could use TimeSpan for those instead of int in milliseconds.

aneojgurhem (approved):
LGTM

dbrasseur-aneo (approved):
Lgtm, but is the Grpc.Core channel the only one needing that ?

lemaitre-aneo:
> Lgtm, but is the Grpc.Core channel the only one needing that ?

In fact, this fix is *not* for Grpc.Core. I will try to do that in another PR, but the options are clumsy.

PR #262, opened 2023-05-30
fix: export in public api from angular package

esoubiran-aneo:
I'm currently trying to in use `next` (3.9.0-x) but I've import issue. This PR fix it.


esoubiran-aneo:
@aneojgurhem this PR is important because without it, I can't continue to dev GUI.

aneojgurhem (approved):

PR #307, opened 2023-06-26
feat: add task options to sort in session

esoubiran-aneo:
We want to be able to sort on every field so we need to add task options to sessions.

Using `tasks_common.TaskOptionField` does not work.


esoubiran-aneo (commented):
I need to update this using messages. like in #343

aneojgurhem:
Let's do it in another PR

aneojgurhem (approved):

PR #316, opened 2023-06-27
ci: release edge on tag

dbrasseur-aneo:
If the edge package to release should have been a normal release and not an edge package, prevents the publish edge from publishing

### DIFF
--- a/.github/workflows/publish-edge.yml
+++ b/.github/workflows/publish-edge.yml
@@ -13,16 +13,20 @@ jobs:
     if: github.actor != 'renovate[bot]' && github.repository_owner == 'aneoconsulting'
     runs-on: ubuntu-latest
     name: Version
+    outputs:
+      version: ${{ steps.genver.outputs.version }}
     steps:
       - name: Checkout
         uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3
         with:
           ref: ${{ github.head_ref }}
           fetch-depth: 0
-
       - name: Generate Version
+        id: genver
+        run: echo "VERSION=$(npx @aneoconsultingfr/generate-next-version@latest --edge)" >> "$GITHUB_OUTPUT"
+      - name: Summary version
       # Print to the summary
-        run: echo "VERSION=$(npx @aneoconsultingfr/generate-next-version@latest --edge)" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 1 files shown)


esoubiran-aneo:
I'm not sur to understand the issue resolved by this PR.

dbrasseur-aneo:
> I'm not sure to understand the issue resolved by this PR @dbrasseur-aneo .

Preventing this from happening again https://github.com/aneoconsulting/ArmoniK.Api/actions/runs/5389726555

esoubiran-aneo:
> > I'm not sure to understand the issue resolved by this PR @dbrasseur-aneo .
> 
> Preventing this from happening again [aneoconsulting/ArmoniK.Api/actions/runs/5389726555](https://github.com/aneoconsulting/ArmoniK.Api/actions/runs/5389726555)

That's a commit on main so a release on main is something normal. I know that's this commit is made for a release but that's still a commit on main. Are we sure we want to add this complexity to check if there is a '-' in the version. Otherwire PR is correct.

dbrasseur-aneo:
> > > I'm not sure to understand the issue resolved by this PR @dbrasseur-aneo .
> > 
> > 
> > Preventing this from happening again [aneoconsulting/ArmoniK.Api/actions/runs/5389726555](https://github.com/aneoconsulting/ArmoniK.Api/actions/runs/5389726555)
> 
> That's a commit on main so a release on main is something normal. I know that's this commit is made for a release but that's still a commit on main. Are we sure we want to add this complexity to check if there is a '-' in the version. Otherwire PR is correct.

Yes ! Because it created a full on release instead of an edga package ! Because we have a 3.9.0 and a 3.9.1 that are identical because of this problem

esoubiran-aneo:
For the python release? for the other packages, it's based on a tag.

dbrasseur-aneo:
> For the python release? for the other packages, it's based on a tag.

No even for csharp

esoubiran-aneo:
> > For the python release? for the other packages, it's based on a tag.
> 
> No even for csharp

It could be an issue in the actions because of the usage of head_ref. We've talk with @aneojgurhem about this.

aneojgurhem (commented):
I'm OK with the generation of the version once as this will prevent to produce inconsistent version if commits/tags are pushed between two steps.
However, the check can be performed once at the end of the version generation step. Everything depending on the generation step will not run. That is enough for what you want to achieve

aneojgurhem (approved):

PR #350, opened 2023-07-11
fix: Implemented Json configuration, fixed includes again

dbrasseur-aneo:


ddubuc (approved):
Just be sure when we do make clean or distclean that simdConfiguration are erased from disk

PR #361, opened 2023-07-21
ci: add a way to release angular package on PR

esoubiran-aneo:


.github/workflows/release-angular-pr.yml
```diff
@@ -0,0 +1,69 @@
+name: Release Angular on PR
```
aneojgurhem, line 1: Should not you put that in edge release ?

aneojgurhem: That's what is made for

esoubiran-aneo: Nope since we run this on every comments

aneojgurhem: a tag in the commit is not enough ? or publishing it every time like the other packages ?

PR #380, opened 2023-08-11
feat: Rust

lemaitre-aneo:
API in Rust


aneojgurhem (commented):
remember to setup `cargo fmt`

packages/rust/armonik/src/client/agent.rs
```diff
@@ -0,0 +1,402 @@
...
+
+    pub async fn create_tasks(
+        &mut self,
+        request: impl Stream<Item = create_tasks::Request> + Send + 'static,
+    ) -> Result<Vec<create_tasks::Status>, super::RequestError> {
+        let response = self.call(request).await?;
+
+        match response {
+            create_tasks::Response::Status {
+                communication_token: _,
```
aneojgurhem, line 101: I think we need a token here

packages/rust/armonik/src/client/sessions.rs
```diff
@@ -0,0 +1,546 @@
...
+            })
+            .await?
+            .session)
+    }
+
+    /// Close a session by its id.
+    pub async fn close(
+        &mut self,
+        session_id: impl Into<String>,
+    ) -> Result<Raw, super::RequestError> {
```
aneojgurhem, line 111: do we keep Raw or RawSession ?

packages/rust/armonik/src/client/events.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub fn with_channel(channel: T) -> Self {
+        Self {
+            inner: v3::events::events_client::EventsClient::new(channel),
+        }
+    }
+
+    /// Get current user
+    pub async fn subscribe(
+        &mut self,
+
```
aneojgurhem, line 32:
```suggestion
```

packages/rust/armonik/examples/client.rs
aneojgurhem, line 1: to me, samples should be in ArmoniK.Samples

packages/rust/armonik/src/client/agent.rs
lemaitre-aneo: No, I don't return the token. Here it is the response.

packages/rust/armonik/src/client/sessions.rs
lemaitre-aneo: The idea was to always use `session::Raw` on the user side. Here, we are in the session module, so it's fine.

lemaitre-aneo:
> Globally, a lot of documentation has not been provided. Do we implement a policy for which rust doc should be present ?

I have not written any doc on the messages that do not have any doc in the protos.

> remember to setup `cargo fmt`

Will do

packages/rust/armonik/examples/client.rs
lemaitre-aneo: sure, it's just that I needed a code to test at some point.

aneojgurhem (approved):

PR #386, opened 2023-08-18
ci: Added ci package uploads

dbrasseur-aneo:


.github/workflows/ci.yml
```diff
@@ -218,7 +218,7 @@ jobs:
         run: python -m pytest tests --cov=armonik --cov-config=.coveragerc --cov-report=term-missing --cov-report xml:coverage.xml --cov-report html:coverage_report
       
       - name: Get Cover 
-        uses: orgoro/coverage@d77626a5fa35d39123e86d6c62907fabe2491496 # v3
+        uses: orgoro/coverage@v3.1
```
aneojgurhem, line 221: why ? this commit should be the latest 3.x

.github/workflows/ci.yml
```diff
@@ -236,9 +236,12 @@ jobs:
           path: packages/python/coverage.xml
 
   build-cpp-packages:
-    name: Build C++ Packages
```
aneojgurhem, line 239: why remove the name ?

.github/workflows/ci.yml
```diff
@@ -236,9 +236,12 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        type: [tar.gz, deb, rpm]
```
aneojgurhem, line 242: I think that a matrix here is not need. You compile 3 times the same thing. Once should be enough

.github/workflows/publish-edge.yml
```diff
@@ -232,3 +232,29 @@ jobs:
           password: ${{ secrets.PYPI_API_TOKEN }}
           print-hash: true
           packages-dir: packages/python/pkg/
+
+    release-cpp-package:
```
aneojgurhem, line 236: You should have some kind of indentation problem here

.github/workflows/ci.yml
dbrasseur-aneo: No they are 3 different dockerfiles with different underlying architectures. tar.gz uses ubuntu 23.04, deb uses debian 12 and rpm uses rhel7

.github/workflows/ci.yml
dbrasseur-aneo: The name is changed a few lines below

.github/workflows/ci.yml
aneojgurhem: Then put the commit hash associated to 3.1

aneojgurhem (approved):

PR #404, opened 2023-08-28
feat: Add filters in events api

aneojgurhem:


Protos/V1/events_common.proto
```diff
@@ -4,14 +4,31 @@ package armonik.api.grpc.v1.events;
...
+  EVENTS_ENUM_NEW_RESULT = 2; /** New restult */
+  EVENTS_ENUM_RESULT_STATUS_UPDATE = 3; /** Result status update */
+  EVENTS_ENUM_RESULT_OWNER_UPDATE = 4; /** Result owner update */
+}
+
 /**
  * Request to subscribe to the event stream.
  */
 message EventSubscriptionRequest {
   string session_id = 1; /** Id of the session that will be used to subscribe events for. **/
```
aneojgurhem, line 28: @lemaitre-aneo do we keep the session Id and put it in the filter internally or do we remove it ?

lemaitre-aneo: I'm not sure. It could be embedded into the filters, but this would allow to listen on everything! I think we should keep this tied to a session.

Protos/V1/events_common.proto
```diff
@@ -3,15 +3,31 @@ syntax = "proto3";
...
+import "results_filters.proto";
 import "task_status.proto";
+import "tasks_filters.proto";
 
 option csharp_namespace = "ArmoniK.Api.gRPC.V1.Events";
 
+/**
+ * Represents the events that can be returned in the EventSubscriptionResponse
+ */
+enum EventsEnum {
```
lemaitre-aneo, line 15: From what we've experienced in Angular, I think we should avoid having naked enums used directly. I would recommend to wrap it into its own message before using it.

Protos/V1/events_common.proto
```diff
@@ -3,15 +3,31 @@ syntax = "proto3";
...
+  EVENTS_ENUM_NEW_TASK = 0; /** New task */
```
lemaitre-aneo, line 16: In Protobuf,iIt is recommended to have an "unspecified" filed that is invalid.

Protos/V1/events_common.proto
aneojgurhem: It depends on where it is used. This is not a case that cause problems. @esoubiran-aneo your thought ?

Protos/V1/events_common.proto
aneojgurhem: added !

Protos/V1/events_common.proto
aneojgurhem: IIRC, there are issues when you put enums directly as part of oneofs. In this case, you have to wrap them into messages

esoubiran-aneo: Yes, works fine in angular. (issue comes when wrapped in a message)

lemaitre-aneo (approved):

PR #407, opened 2023-09-04
feat: new overloads for worker server creation

aneojgurhem:


lemaitre-aneo (commented):
What is the purpose of this?

packages/csharp/ArmoniK.Api.Worker/Utils/WorkerServer.cs
```diff
@@ -48,6 +48,18 @@ namespace ArmoniK.Api.Worker.Utils;
...
+  /// <summary>
+  ///   Create a web application for the given ArmoniK Worker gRPC Service
+  /// </summary>
+  /// <typeparam name="T">gRPC Service to add to the web application</typeparam>
+  /// <returns>
+  ///   The web application initialized
+  /// </returns>
+  public static WebApplication Create<T>()
+    where T : gRPC.V1.Worker.Worker.WorkerBase
+    => Create<T>(null);
```
lemaitre-aneo, line 60: This should be handled by the following function:

```c#
  public static WebApplication Create<T>(IConfiguration?             configuration       = null,
                                         Action<IServiceCollection>? serviceConfigurator = null)
```

lemaitre-aneo: So you should most likely remove the optional first parameter to the other method.

aneojgurhem: but it breaks backward compatibility

lemaitre-aneo: It shouldn't. Default parameters are handled by the caller, not the callee. Have you tried?

aneojgurhem: Yes. That's why I have the issue.  There is an unit test for that. You can try to remove the empty parameter overload and compile

lemaitre-aneo (approved):

PR #413, opened 2023-09-13
feat: C++14 support and various installation fixes

dbrasseur-aneo:
- Option to change url of dependencies
- Cpp14 support
- Explicit namespaces
- Fix protobuf package issue on windows


packages/cpp/ArmoniK.Api.Client/source/submitter/ResultsClient.cpp
```diff
@@ -64,21 +66,23 @@ void ResultsClient::upload_result_data(const std::string &session_id, const std:
 
   while (offset < payload.size()) {
     size_t chunkSize = std::min(maxChunkSize, payload.size() - offset);
```
lemaitre-aneo, line 68: you could simplify this like this:
```suggestion
  while (payload.size()) {
    auto chunk = payload.substr(0, maxChunkSize);
    payload = payload.substr(maxChunkSize);
```

packages/cpp/ArmoniK.Api.Worker/source/Worker/TaskHandler.cpp
```diff
@@ -55,11 +54,9 @@ void API_WORKER_NAMESPACE::TaskHandler::init() {
   token_ = Request.communication_token();
   config_ = std::move(*init_request->mutable_configuration());
 
-  std::vector<std::string> chunks;
-
   auto *datachunk = &init_request->payload();
-  payload_.resize(datachunk->data().size());
-  std::memcpy(payload_.data(), datachunk->data().data(), datachunk->data().size());
+  payload_.clear();
```
lemaitre-aneo, line 58: I would prefer an assert instead

lemaitre-aneo (dismissed):
Good enough for me

packages/cpp/ArmoniK.Api.Common/header/logger/util.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+#include <absl/strings/string_view.h>
+#include <fmt/std.h>
+
+const fmt::string_view to_fmt(const absl::string_view sv) { return fmt::string_view(sv.data(), sv.size()); }
```
ddiakiteaneo, line 6:
```suggestion
const fmt::string_view to_fmt(const absl::string_view sv) { return fmt::string_view(sv.data(), sv.size()); }

```

packages/cpp/README.md
```diff
@@ -0,0 +1,17 @@
+# ArmoniK Api Cpp
+
+## Build requirements
+- cmake 3.22+
+- C++ compiler with C++14 support
+- grpc 1.54 - 1.56.2
```
ddiakiteaneo, line 6: It can work if grpc version is > 1.50 I think, but it's ok

packages/cpp/ArmoniK.Api.Client/source/submitter/ResultsClient.cpp
```diff
@@ -21,7 +23,7 @@ std::map<std::string, std::string> ResultsClient::create_results(std::string_vie
   }
 
   results_request.mutable_results()->Add(results_create.begin(), results_create.end());
-  *results_request.mutable_session_id() = session_id;
+  results_request.mutable_session_id()->assign(session_id.data(), session_id.size());
```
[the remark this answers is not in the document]
dbrasseur-aneo: It simply doesn't compile in C++14

ddiakiteaneo (approved):

PR #415, opened 2023-09-18
fix: bad chunk size for result upload

dbrasseur-aneo:


packages/cpp/ArmoniK.Api.Tests/source/SubmitterClientTest.cpp
```diff
@@ -253,26 +256,31 @@ TEST(testMock, getResult) {
...
   armonik::api::grpc::v1::ResultRequest result_request;
 
-  init(channel, task_options);
+  init(channel, task_options, log);
 
   auto stub = armonik::api::grpc::v1::results::Results::NewStub(channel);
 
   grpc::ClientContext context;
 
+  log.info("Creating Client");
```
aneojgurhem, line 265: I think this kind of log should be debug level

dbrasseur-aneo: done

ngruelaneo (approved):

PR #416, opened 2023-09-20
feat: Add Task List and filters to Python

dbrasseur-aneo:


packages/common/protofiles.sh
```diff
@@ -5,17 +5,5 @@ export PROTO_PATH=$REPOSITORY_PATH/Protos/V1
...
-                      "results_common.proto" "applications_common.proto" "auth_common.proto" \
-                      "events_common.proto" "partitions_common.proto" "sort_direction.proto" \
-                      "versions_common.proto" "tasks_fields.proto" "tasks_filters.proto" \
-                      "sessions_fields.proto" "sessions_filters.proto" \
-                      "applications_fields.proto" "applications_filters.proto" \
-                      "partitions_fields.proto" "partitions_filters.proto" \
-                      "results_fields.proto" "results_filters.proto" \
-                      "filters_common.proto")
+if IFS=$'\n' read -rd '' -a armonik_client_files <<<"$(find $PROTO_PATH -name "*.proto" -exec basename {} \; | grep -v -e "worker" -e "agent" | grep "service" )"; then :; fi
+if IFS=$'\n' read -rd '' -a armonik_common_files <<<"$(find $PROTO_PATH -name "*.proto" -exec basename {} \; | grep -v "service" )"; then :; fi
```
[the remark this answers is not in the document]
aneojgurhem: That's not an issue. We just have to update it we we miss something

packages/python/src/armonik/client/submitter.py
aneojgurhem, line 1: I would not put any effort on this file as it will be deleted soon.

dbrasseur-aneo: Soon yes, but the other apis haven't been "pythonified" yet

packages/python/src/armonik/common/filter.py
```diff
@@ -0,0 +1,443 @@
...
+from google.protobuf.message import Message
+import google.protobuf.timestamp_pb2 as timestamp
+from ..protogen.common.filters_common_pb2 import *
+
+
+class Filter:
+    """
+    Base class for all filters
+    """
+    def message_type(self) -> Type:
```
aneojgurhem, line 12: Use proper decorators and classes to say this is an abstract class. Do not re-implement the wheel

packages/python/src/armonik/common/filter.py
```diff
@@ -0,0 +1,443 @@
...
+        return raw
+
+    def to_conjunction(self) -> "FilterConjunction":
+        raise Exception("Cannot transform a disjunction into a conjunction")
+
+    def to_disjunction(self) -> "FilterDisjunction":
+        return self
+
+
+class FilterConjunction(Filter):
```
aneojgurhem, line 112: I think these should not inherit `Filter`

aneojgurhem: They are at a higher level in the AST

dbrasseur-aneo: Filter is more of an interface in this case. So it wouldn't be that strange imo.

aneojgurhem (approved):

packages/python/src/armonik/common/filter.py
```diff
@@ -0,0 +1,327 @@
...
+            new filter with the given value/operator
+
+        Raises:
+            NotImplementedError if the given operator is not available for the given class
+        """
+        if self.is_true_conjunction() or self.is_true_disjunction():
+            raise Exception(f"Cannot apply operator to a disjunction or a conjunction")
+        self._verify_value(value)
+        if operator is None:
+            raise NotImplementedError(f"Operator {operator_str} is not available for {self.__class__.__name__}")
```
ngruelaneo, line 199: same no fstring

ngruelaneo (changes requested):

ngruelaneo (approved):

PR #421, opened 2023-09-26
fix: Agent/Worker comm using files in Python and Cpp

dbrasseur-aneo:


dbrasseur-aneo:
Cla bot not giving the comment...

packages/cpp/ArmoniK.Api.Worker/header/Worker/TaskHandler.h
```diff
@@ -89,7 +84,8 @@ class TaskHandler {
    * @param data The result data
    * @return A future containing a vector of ResultReply
    */
-  std::future<armonik::api::grpc::v1::agent::ResultReply> send_result(std::string key, absl::string_view data);
+  std::future<armonik::api::grpc::v1::agent::NotifyResultDataResponse> send_result(std::string key,
```
aneojgurhem, line 87: It is unfortunate that it changes the API. I think you should wrap this message somehow

packages/cpp/tools/restart-build-env.sh
aneojgurhem, line 1: why does this change in this PR ?

packages/cpp/ArmoniK.Api.Worker/header/Worker/TaskHandler.h
aneojgurhem: I'd return nothing and throw if there is an issue

aneojgurhem (approved):

PR #433, opened 2023-10-18
feat: Add Result List with a Status Filter to Python

jeremyzyngeraneo:


aneojgurhem (approved):

packages/python/src/armonik/common/objects.py
```diff
@@ -182,3 +183,29 @@ def from_message(cls, session_raw: SessionRaw) -> "Session":
             cancelled_at=timestamp_to_datetime(session_raw.cancelled_at),
             duration=duration_to_timedelta(session_raw.duration)
         )
+
+from dataclasses import dataclass, field
+from typing import List, Optional
+from datetime import datetime
```
dbrasseur-aneo, line 189: Imports should be a the top of the file

dbrasseur-aneo (approved):
Not a big issue, we can merge as is

PR #436, opened 2023-10-25
refactor: Task service wrapper cpp

dbrasseur-aneo:
- To be rebased once #435 is merged

### DIFF
--- a/packages/cpp/ArmoniK.Api.Common/header/objects/Task.h
+++ b/packages/cpp/ArmoniK.Api.Common/header/objects/Task.h
@@ -6,7 +6,7 @@
 
 namespace armonik {
 namespace api {
-namespace client {
+namespace common {
 struct TaskCreation {
   /**
    * Payload Id
@@ -26,7 +26,6 @@ struct TaskCreation {
    * Per task task options, none by default
    */
   armonik::api::grpc::v1::TaskOptions taskOptions = get_no_task_options();
-
 
   /**
    * Default "no task option" value
@@ -60,6 +59,6 @@ struct TaskInfo {
    */
   std::string payload_id;
 };
-} // namespace client
+} // namespace common
 } // namespace api
 } // namespace armonik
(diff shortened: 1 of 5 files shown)


packages/cpp/ArmoniK.Api.Common/header/objects/Task.h
```diff
@@ -0,0 +1,64 @@
...
+   */
+  armonik::api::grpc::v1::TaskOptions taskOptions = get_no_task_options();
+
+  /**
+   * Default "no task option" value
+   * @return A task option to use to ignore the task options parameter
+   */
+  static armonik::api::grpc::v1::TaskOptions get_no_task_options() {
+    armonik::api::grpc::v1::TaskOptions options;
+    options.set_max_retries(INT32_MIN);
```
aneojgurhem, line 36: Is that signed ?

aneojgurhem: OK so the min is omething negative. I would not advise a negative number of retries. 1 as default is good enough

dbrasseur-aneo: The idea here is not to have a "proper default". It's to detect whether the user actually put a TaskOption or not.
The rationale is that we don't want to put explicitely a TaskOption in the request (because otherwise it would be merged by core). The reason for not using the default value is that it can then distinguish between someone who put a task option but forgot to set the max_retries value, and someone who doesn't want to put an overriding task option.

Another solution could be a pointer, but I'm not a fan of leaving pointers in pod structures

aneojgurhem: null is not enough ?

aneojgurhem (dismissed):

packages/cpp/ArmoniK.Api.Common/header/objects/Task.h
dbrasseur-aneo: Iif you have a pointer yes, otherwise you need some sort of mechanism to detect whether the user has put an actual value

aneojgurhem (approved):

PR #441, opened 2023-11-22
docs: documentation and example for grpc secure channel with python

jeremyzyngeraneo:


.docs/content/2.usage/use-armonik-grpc-securely-python.md
```diff
@@ -0,0 +1,75 @@
...
+sudo nano /etc/hosts
+```
+
+### Use ArmoniK Endpoint in Python
+
+Use `armonik.local` as endpoint and don't forget to specify the port
+
+```bash
+armonik.local:5001
+```
```
dbrasseur-aneo, line 49: This is true only : 
- If the given certificate common name doesn't match the endpoint name
- armonik.local is just the default common name

.docs/content/2.usage/use-armonik-grpc-securely-python.md
```diff
@@ -0,0 +1,75 @@
...
+    ```hcl
+    tls = true
+    mtls = true
+    ```
+
+2. Deploy ArmoniK.
+
+3. Certificates will be generated in `all-in-one/generated/certificates/ingress`:
+   - For TLS: `ca.crt`
+   - For mTLS: `ca.crt`, `client.submitter.crt`, `client.submitter.key`
```
dbrasseur-aneo, line 31: ArmoniK.Api doesn't depend on the infrastructure. So I'm unsure if this documentation should be here ? @aneojgurhem what do you think ?

examples/python/secure_grpc_client.py
```diff
@@ -0,0 +1,59 @@
...
+    Args:
+        ca (str): CA file path for TLS or mutual TLS
+        cert (str): Certificate file path for mutual TLS
+        key (str): Private key file path for mutual TLS
+        endpoint (str): ArmoniK control plane endpoint
+
+    Returns:
+        grpc.Channel: gRPC channel for communication
+    """
+    if ca != None:
```
dbrasseur-aneo, line 28: The CA is only mandatory if you've not installed it on your system. cf https://ubuntu.com/server/docs/security-trust-store

What you should test for is whether the endpoint is called using http or https (or if it doesn't contain the scheme, then by a flag (-ssl for example))
In the case you want simple tls without a ca, you can construct a default empty grpc.ssl_channel_credentials

examples/python/secure_grpc_client.py
```diff
@@ -0,0 +1,59 @@
...
+            with open(ca, 'rb') as ca_file:
+                ca_data = ca_file.read()
+            if cert != None and key!= None :
```
dbrasseur-aneo, line 31:
```suggestion
            if cert is not None and key is not None :
```
None checks should be done with "is" and "is not" instead of equal https://peps.python.org/pep-0008/#programming-recommendations

examples/python/secure_grpc_client.py
```diff
@@ -0,0 +1,59 @@
...
+    # Create the session client
+    client = ArmoniKSessions(channel)
+    # List numbers sessions with a cancelled status filter
+    sessions = client.list_sessions(SessionFieldFilter.STATUS == SESSION_STATUS_CANCELLED)
+        
+    print(f'\nNumber of sessions: {sessions[0]}\n')
+
+
+if __name__ == "__main__":
+    main()
```
dbrasseur-aneo, line 59:
```suggestion
    main()

```
Missing empty last line

examples/python/secure_grpc_client.py
```diff
@@ -0,0 +1,59 @@
...
+from armonik.client.sessions import ArmoniKSessions, SessionFieldFilter
+from armonik.common.enumwrapper import SESSION_STATUS_CANCELLED
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(description="ArmoniK Example for Client connection TLS or mutual TLS")
+    parser.add_argument("-v","--version", action="version", version="ArmoniK Admin CLI 0.0.1")
+    parser.add_argument("--endpoint", default="localhost:5001", help="ArmoniK control plane endpoint")
+    parser.add_argument("--ca", help="ca.crt path for TLS or mutual TLS")
+    parser.add_argument("--cert", help="client.submitter.crt for mutual TLS")
+    parser.add_argument("--key", help="client.submitter.key for mutual TLS")
```
dbrasseur-aneo, line 12:
```suggestion
    parser.add_argument("--cert", help="client certificate path for mutual TLS")
    parser.add_argument("--key", help="client key path for mutual TLS")
```

dbrasseur-aneo (changes requested):

dbrasseur-aneo (approved):

PR #460, opened 2024-01-04
refactor: replace click with arparse

jeremyzyngeraneo:


ngruelaneo:
enelve aussi la dependance dans le fichier proto2python.sh (pip install .... click ... )

aneojgurhem (dismissed):

packages/python/fix_imports.py
```diff
@@ -11,19 +11,13 @@
...
 class ProtobufFilePathInfo(TypedDict):
     dir: Path
     path: Path
     rel_path: Path
 
-
-@click.command()
-@click.option("--dry", is_flag=True, show_default=True, default=False,
-              help="Do not write out the changes to the files.")
-@click.argument("root_dir", type=click.Path(exists=True))
```
qdelamea-aneo, line 26: As stated in the doc "The Path type has several checks available which raise nice errors if they fail, such as existence." (see https://click.palletsprojects.com/en/8.1.x/arguments/). It means that the root_dir must be an existing directroy and that an error is raised if not. This behavior was implicitly implemented by click and now must be re-implemented after argument parsing.

aneojgurhem (dismissed):

packages/python/fix_imports.py
```diff
@@ -11,25 +11,22 @@
...
 def fix_protobuf_imports(root_dir, dry):
     """
       A script to fix relative imports (from and to nested sub-directories) within compiled `*_pb2*.py` Protobuf files.
     """
 
     root_dir = Path(root_dir)
+    if not root_dir.is_dir():
+        print(f"Error: The specified root directory '{root_dir}' does not exist.")
```
qdelamea-aneo, line 28: It is better to raise a suitable exception instead of printing. Try something like:
````
raise argparse.ArgumentError("rout_dir", "<error message>")
```

packages/python/fix_imports.py
```diff
@@ -11,25 +11,22 @@
...
+        sys.exit(1)
```
qdelamea-aneo, line 29: Raise exception and then delete this line.

ngruelaneo (changes requested):
add test on directory existence

packages/python/fix_imports.py
```diff
@@ -142,7 +136,11 @@ def fix_protobuf_imports_in_file(name, info: ProtobufFilePathInfo, pyi=False):
 
 
 def main():
-    fix_protobuf_imports()
+    parser = argparse.ArgumentParser()
+    parser.add_argument("root_dir", type=str, help="Path to the root directory")
+    parser.add_argument("--dry", action="store_true", default=False, help="Do not write out the changes to the files.")
+    args = parser.parse_args()
+    fix_protobuf_imports(args.root_dir, args.dry)
```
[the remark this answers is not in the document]
ngruelaneo: je n'avais pas lu la doc de argparse qui fourni aussi des outils de tests :)

packages/python/proto2python.sh
```diff
@@ -33,7 +33,7 @@ python -m pip install --upgrade pip
 python -m venv $PYTHON_VENV
 source $PYTHON_VENV/bin/activate
 # We need to fix grpc to 1.56 until this bug is solved : https://github.com/grpc/grpc/issues/34305
-python -m pip install build grpcio==1.56.2 grpcio-tools==1.56.2 click pytest setuptools_scm[toml]
+python -m pip install build grpcio==1.56.2 grpcio-tools==1.56.2 pytest setuptools_scm[toml]
```
ngruelaneo, line 36:
```suggestion
python -m pip install build grpcio==1.56.2 grpcio-tools==1.56.2 setuptools_scm[toml]
```

ngruelaneo (approved):

PR #479, opened 2024-02-14
refactor: improve events client extension to wait for result

aneojgurhem:


ngruelaneo (approved):

packages/csharp/ArmoniK.Api.Client/EventsClientExt.cs
```diff
@@ -80,61 +83,71 @@ private static FiltersAnd ResultsFilter(string resultId)
...
+              {
+                throw new ResultAbortedException($"Result {resp.NewResult.ResultId} has been aborted");
+              }
+            }
           }
         }
+        catch (OperationCanceledException)
+        {
+        }
+        catch (RpcException)
```
lemaitre-aneo, line 148: Maybe a bit too wide?

lemaitre-aneo (approved):

packages/csharp/ArmoniK.Api.Client/EventsClientExt.cs
aneojgurhem: we can improve it as we find things we nned to exclude

PR #495, opened 2024-04-18
fix: Abort immediately upon cancellation

lemaitre-aneo:
Change the default option and abort immediately upon cancellation.

This is to limit the edge case that a task cancellation that is not catched would make the next task to crash and being retried.


aneojgurhem (approved):

lemaitre-aneo:
The proper fix would be to add a rpc that the worker would call to inform the agent that the call has actually been cancelled.

PR #498, opened 2024-04-18
chore: Update to version 3.16.3

ngruelaneo:


lemaitre-aneo:
As there is a feat since 3.16.2, it will be a 3.17.0.
Besides, linting fails

PR #522, opened 2024-06-20
fix: Added winHttpHandler TcpKeepAlive

dbrasseur-aneo:


dbrasseur-aneo:
On hold until we find a solution to detect the compatibility

PR #530, opened 2024-07-17
fix: Python API event client wait_for_result_availability method

qdelamea-aneo:
- fix: Python API event client wait_for_result_availability method
- fix: Python event client

### DIFF
--- a/packages/python/src/armonik/client/events.py
+++ b/packages/python/src/armonik/client/events.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from typing import Callable, cast, Iterable, List, Optional
 
 from grpc import Channel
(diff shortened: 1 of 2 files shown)


packages/python/src/armonik/client/events.py
```diff
@@ -56,66 +57,84 @@ def get_events(
...
-            request.tasks_filters = cast(rawTaskFilters, task_filter.to_disjunction().to_message())
-        if result_filter:
-            request.results_filters = cast(
-                rawResultFilters, result_filter.to_disjunction().to_message()
-            )
+        request = EventSubscriptionRequest(
+            session_id=session_id,
+            returned_events=event_types,
+            tasks_filters=cast(rawTaskFilters, task_filter.to_disjunction().to_message())
+            if task_filter
```
dbrasseur-aneo, line 71:
```suggestion
            if task_filter is not None
```
Filters to not implement "truthyness" as they can't be converted to booleans. Other instances of this are present in other parts of the file

packages/python/src/armonik/client/events.py
```diff
@@ -56,66 +57,84 @@ def get_events(
...
-    def wait_for_result_availability(self, result_id: str, session_id: str) -> None:
+            for event_handler in event_handlers:
+                if event_handler(
+                    session_id,
+                    EventTypes.from_string(event_type),
+                    self._events_obj_mapping[event_type].from_raw_event(
+                        getattr(message, event_type)
+                    ),
+                ):
+                    break
```
dbrasseur-aneo, line 89: This does not do what the docstring describes.
When an handler return True, it stops processing other handlers, but it doesn't stop listening as the "break" does not break the message loop.
you could use :  
```suggestion
            if all(not event_handler(
                session_id,
                EventTypes.from_string(event_type),
                self._events_obj_mapping[event_type].from_raw_event(
                    getattr(message, event_type)
                ),
            ) for event_handler in event_handlers):
                break
```

packages/python/src/armonik/client/events.py
```diff
@@ -56,66 +57,84 @@ def get_events(
...
+
+    def wait_for_result_availability(self, result_ids: List[str], session_id: str) -> None:
```
dbrasseur-aneo, line 91: I would keep the old one too with a
```suggestion
    def wait_for_result_availability(self, result_ids: Union[str, List[str]], session_id: str) -> None:
```
And wrap the singular one in a list
for Api compatibility

packages/python/src/armonik/client/events.py
```diff
@@ -56,66 +57,84 @@ def get_events(
...
-            return
-        elif result.status == ResultStatus.ABORTED:
-            raise RuntimeError(f"Result {result.name} with ID {result_id} is aborted.")
-
-        self.get_events(
-            session_id,
-            [EventTypes.RESULT_STATUS_UPDATE],
-            [handler],
-            result_filter=(ResultFieldFilter.RESULT_ID == result_id),
+        results_not_found = deepcopy(result_ids)
```
dbrasseur-aneo, line 102: A simple copy should be sufficient, as a string is immutable in Python

packages/python/src/armonik/client/events.py
```diff
@@ -56,66 +57,84 @@ def get_events(
...
+
+        results_filter = ResultFieldFilter.RESULT_ID == result_ids[0]
```
dbrasseur-aneo, line 104: Needs a check for empty list

dbrasseur-aneo (changes requested):

aneojgurhem (approved):

PR #536, opened 2024-08-06
ci: add java package

sboukhetta:
Add java package release

### DIFF
--- a/packages/java/pom.xml
+++ b/packages/java/pom.xml
@@ -7,7 +7,7 @@
   <groupId>fr.aneo</groupId>
   <artifactId>armonik-java</artifactId>
   <description>GRPC java binding for the Armonik orchestrator API</description>
-  <version>0.1.0</version>
+  <version>3.19.0</version>
 
 	<name>armonik-java</name>
 	<url>https://github.com/aneoconsulting/ArmoniK.Api</url>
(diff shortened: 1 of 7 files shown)


packages/java/pom.xml
```diff
@@ -4,11 +4,40 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
-  <groupId>org.aneo</groupId>
+  <groupId>fr.aneo</groupId>
```
ngruelaneo, line 7: perhaps `eu.aneo` should be better. aneo.fr is only a redirection I think.

packages/java/pom.xml
```diff
@@ -4,11 +4,40 @@
...
   <artifactId>armonik-java</artifactId>
   <description>GRPC java binding for the Armonik orchestrator API</description>
   <version>0.1.0</version>
 
+	<name>armonik-java</name>
+	<url>https://github.com/aneoconsulting/ArmoniK.Api</url>
+
+	<developers>
```
ngruelaneo, line 15: I think that we should add Hassan who start it even if he is not anymore with aneo.

packages/java/pom.xml
jdongo-aneo: the namespace has been created with the fr.aneo after discussing with Maxime

packages/java/pom.xml
sboukhetta: Yes, I will add him.

dbrasseur-aneo (approved):

PR #537, opened 2024-08-19
perf: improve wait for results performances by reducing the size of the request and making multiple at the same time

aneojgurhem:


packages/csharp/ArmoniK.Api.Client/EventsClientExt.cs
dbrasseur-aneo, line 1: Please fix xml warnings

packages/csharp/ArmoniK.Api.Client/EventsClientExt.cs
```diff
@@ -80,75 +81,85 @@ private static FiltersAnd ResultsFilter(string resultId)
...
-        catch (RpcException)
-        {
-        }
-      }
-    }
+                                                   if (resp.NewResult.Status == ResultStatus.Aborted)
+                                                   {
+                                                     throw new ResultAbortedException($"Result {resp.NewResult.ResultId} has been aborted");
+                                                   }
+                                                 }
```
dbrasseur-aneo, line 153: You can handle both cases without repetition by extracting the "status" and "resultId" first depending on the event type, then do the checks and throw if necessary

dbrasseur-aneo (changes requested):

dbrasseur-aneo (approved):

PR #542, opened 2024-08-26
chore: Move Python examples

dbrasseur-aneo:


aneojgurhem (dismissed):
Should be in samples. Maybe time to move it directly in samples

dbrasseur-aneo:
> Should be in samples. Maybe time to move it directly in samples

Let's merge it, and we'll move it out after 3.20

aneojgurhem:
> > Should be in samples. Maybe time to move it directly in samples
> 
> Let's merge it, and we'll move it out after 3.20 or do we get it out now ?

In my opinion, it should be now.

aneojgurhem (approved):

PR #546, opened 2024-08-29
feat: Add tls mtls support for c++ api

ddiakiteaneo:


packages/cpp/ArmoniK.Api.Client/header/channel/ChannelFactory.h
```diff
@@ -0,0 +1,32 @@
...
+  std::shared_ptr<::grpc::Channel> create_channel();
+
+  static bool ShutdownOnFailure(std::shared_ptr<::grpc::Channel> channel);
+
+private:
+  armonik::api::common::logger::LocalLogger logger_;
+  std::shared_ptr<::grpc::ChannelCredentials> credentials_{nullptr};
+  std::string endpoint_;
+  armonik::api::common::utils::Configuration configuration_;
+  bool is_secure_{false};
```
dbrasseur-aneo, line 28: Is this used for anything ?

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
```diff
@@ -0,0 +1,165 @@
...
+      }
+    } else {
+      if (control_plane.isSslValidation()) {
+        credentials_ = grpc::SslCredentials(grpc::SslCredentialsOptions{std::move(root_cert_pem)});
+      } else {
+        TlsChannelCredentialsOptions tls_options;
+        tls_options.set_certificate_provider(
+            create_certificate_provider(root_self_signed, user_public_pem, user_private_pem));
+        tls_options.set_verify_server_certs(control_plane.isSslValidation());
+        credentials_ = TlsCredentials(tls_options);
```
lemaitre-aneo, line 123: I thought you could not disable ssl validation

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
```diff
@@ -0,0 +1,165 @@
...
+      }
+    }
+    is_secure_ = true;
+  } else {
+    credentials_ = grpc::InsecureChannelCredentials();
```
lemaitre-aneo, line 128: Does this still works?

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+  if (channel != nullptr) {
+    if (ShutdownOnFailure(channel)) {
+      logger_.log(common::logger::Level::Debug, "Shutdown unhealthy channel");
+    } else {
+      logger_.log(common::logger::Level::Debug, "Valid channel");
+    }
+  }
+  return channel;
+}
```
dbrasseur-aneo, line 96: Why ? You just created it so it's not null. Shutdown on failure is used in channel pools, but not for this kind of factory

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
```diff
@@ -0,0 +1,165 @@
...
+        credentials_ = TlsCredentials(tls_options);
+      }
+    }
+    is_secure_ = true;
+  } else {
+    credentials_ = grpc::InsecureChannelCredentials();
+  }
+}
+
+bool ChannelFactory::ShutdownOnFailure(std::shared_ptr<grpc::Channel> channel) {
```
dbrasseur-aneo, line 132: Not sure it's the right place for it, as it's more for a channel pool

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
ddiakiteaneo: It can not be disabled only in the case where the client certificate and key are specified

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
ddiakiteaneo: Yes it works because the secure lib of grpc (gRPC::grpc++) supports both secure and unsecure connexion

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
ddiakiteaneo: I used the "Shutdown on failure" because when debugging I noticed that the created channel can be in failure that's why I added the check after the creation.

lemaitre-aneo: I think the comment was more about `if (channel != nullptr)`

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
lemaitre-aneo: Ok. That's very specific, but I'm actually glad that's only for this very special case that won't happen in prod, and should not be the norm for devs.

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
lemaitre-aneo: I don't see any issue to provide it in API as a utility function.
But signature is odd. You should take `std::shared_ptr` only if you need ownership of it, but here, that's not the case. You can go with a plain reference.

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include <sstream>
+#include <utility>
+
+namespace armonik {
+namespace api {
+namespace client {
+
+using namespace grpc::experimental;
+
+const std::string root_self_signed = R"(-----BEGIN CERTIFICATE REQUEST-----
```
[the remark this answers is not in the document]
lemaitre-aneo: I agree it needs documentation. If I recall correctly, disabling the ssl validation still requires to pass a root cert. So this is just a dummy certificate to pass to the function.
It would also probably need the command used to generate it in case it needs to be recreated for whatever reason.

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
```diff
@@ -56,17 +65,47 @@ std::string get_key(const absl::string_view &path) {
...
-    endpoint_view = endpoint_view.substr(0, delim);
+  const auto http_delim = endpoint_view.find("http://");
+  const auto https_delim = endpoint_view.find("https://");
+  if (endpoint_view.find("unix") == 0) {
+    if (endpoint_view[0] == '/') {
+      endpoint.insert(0, "unix://");
+    } else {
+      endpoint.insert(0, "unix:");
+    }
+    return false;
```
dbrasseur-aneo, line 85: If "unix" is at index 0, then "/" cannot be at index 0 :p

packages/cpp/ArmoniK.Api.Client/source/channel/ChannelFactory.cpp
```diff
@@ -45,7 +49,12 @@ bDCNVTg3w/OQLQQdWUl6FunmYinukBgmqnsJnwgrhzBENbmgbgfOZZWGtG5ODENb
 wc+KqiSg9c9iqA==
 -----END CERTIFICATE REQUEST-----)";
 
-std::string get_key(const absl::string_view &path) {
+/**
+ *
+ * @param path
+ * @return
+ */
+std::string read_file(const absl::string_view &path) {
```
dbrasseur-aneo, line 57: Empty doc

packages/cpp/ArmoniK.Api.Client/header/channel/ChannelFactory.h
ddiakiteaneo: Not now, but can be used in tests to check if the channel is secure or not

dbrasseur-aneo (approved):

Issue #548, opened 2024-08-30
[PY] Detailed = False when listing tasks not working


Pyrrha:
Hello,

I'm using armonik==3.17.0. In this version, when I'm retrieving tasks without using `detailed` option, I get the task result. However, when using it `detailed = False`, I get the following error:

```
   ...
    result = client.list_tasks(
             ^^^^^^^^^^^^^^^^^^
  File "/mnt/c/Users/x/Documents/project/venv/lib/python3.12/site-packages/armonik/client/tasks.py", line 333, in list_tasks
    return response.total, [Task.from_message(t) for t in response.tasks]
                            ^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/Users/x/Documents/project/venv/lib/python3.12/site-packages/armonik/common/objects.py", line 157, in from_message
    parent_task_ids=list(task_raw.parent_task_ids),
                         ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: parent_task_ids
```

Does it requires another parameter to be set in order to work?

Thanks

### Versions used

* ArmoniK SDK : 3.17.0
* Core : 0.24.4.0
* API : 3.19.0.0

dbrasseur-aneo:
Hi,
You're right, there is a bug there.
It will be corrected in 3.20.
In the meantime, please use only the detailed version.

Best regards,
Dylan

PR #549, opened 2024-08-30
fix: task from task summary

dbrasseur-aneo:
fixes: #548


lemaitre-aneo:
I think you should go the other way around: construct a summary from a detailed.

aneojgurhem (approved):

PR #550, opened 2024-08-30
feat: armonik_worker decorator

dbrasseur-aneo:


packages/python/src/armonik/worker/worker.py
```diff
@@ -74,3 +79,87 @@ def Process(self, request: ProcessRequest, context) -> Union[ProcessReply, None]
...
         return HealthCheckReply(status=self.health_check())
+
+
+def armonik_worker(
+    *,
+    autorun: bool = False,
+    logger: Optional[Logger] = None,
+    worker_endpoint: Optional[str] = None,
+    agent_endpoint: Optional[str] = None,
+    channel_options: Tuple[Tuple[str, str]] = (("grpc.default_authority", "localhost"),),
```
lemaitre-aneo, line 90: I would argue this should not be arguments of the decorator, but arguments to the run function.
Also, it might be simpler to construct an object instead of nesting functions like that.

packages/python/src/armonik/worker/worker.py
```diff
@@ -68,9 +73,119 @@ def Process(self, request: ProcessRequest, context) -> Union[ProcessReply, None]
...
+
+class ArmoniKWorkerWrapper:
+    def __init__(
+        self,
+        *,
+        processor: Callable[[TaskHandler], Output],
+        logger: Optional[Logger] = None,
+        worker_endpoint: Optional[str] = None,
+        agent_endpoint: Optional[str] = None,
+        channel_options: Optional[Tuple[Tuple[str, str]]] = _NOT_SET,
```
lemaitre-aneo, line 95: Why using `_NOT_SET` instead of `None`?

packages/python/src/armonik/worker/worker.py
```diff
@@ -68,9 +73,119 @@ def Process(self, request: ProcessRequest, context) -> Union[ProcessReply, None]
...
+            )
+        if channel_options is _NOT_SET:
+            channel_options = (("grpc.default_authority", "localhost"),)
+        self.logger = logger
+        self.worker_endpoint = worker_endpoint
+        self.agent_endpoint = agent_endpoint
+        self.channel_options = channel_options
+        self.processor = processor
+
+    def __call__(self, channel: Optional[Channel] = None):
```
lemaitre-aneo, line 128: I would say this function should be `run`, and let `__call__` just be a call to `self.processor` unwrapped.

packages/python/src/armonik/worker/worker.py
```diff
@@ -68,9 +73,119 @@ def Process(self, request: ProcessRequest, context) -> Union[ProcessReply, None]
...
+            self.logger.info("Worker Connected")
+            worker.start(self.worker_endpoint)
+        except Exception as e:
+            if channel is None:
+                channel_.__exit__(type(e), e, e.__traceback__)
+                channel_ = None
+            raise
+        finally:
+            if channel is None and channel_ is not None:
+                channel_.__exit__(None, None, None)
```
lemaitre-aneo, line 147: I am not sure why it could not be a `with` block

dbrasseur-aneo: An external channel should not be reopened and closed inside, as its lifetime is handled outside the function

lemaitre-aneo (approved):

PR #554, opened 2024-09-03
test: Adapt C++ tests to server mock

ddiakiteaneo:


packages/cpp/ArmoniK.Api.Tests/CMakeLists.txt
dbrasseur-aneo, line 1: ArmoniK.Api.Common already has the simdjson library to handle json

dbrasseur-aneo (approved):

PR #570, opened 2024-11-18
feat: turn enumwrappers into real python enumerators

qdelamea-aneo:
- feat: turn enumwrappers into real python enumerators
- fix: correctly convert TaskOptions option property into Python dictionnary

# Motivation

So far, the API wrappers for enumerators are not “real” Python enumerators. This makes it difficult to obtain the name associated with the value of enumerator elements.

# Description

This change fixes this problem by making each enumerator wrapper a “real” Python enumerator.

It also fixes a minor bug in the conversion of the TaskOptions option field to a native Python dictionary.

# Testing

A few manual tests in addition to unit tests.

# Impact

This changes should have no impact.

# Additional Information

# Checklist

### DIFF
--- a/packages/python/src/armonik/common/enumwrapper.py
+++ b/packages/python/src/armonik/common/enumwrapper.py
@@ -75,7 +75,7 @@ class TaskStatus(IntEnum):
     PAUSED = TASK_STATUS_PAUSED
 
 
-class EventTypes:
+class EventTypes(IntEnum):
     UNSPECIFIED = EVENTS_ENUM_UNSPECIFIED
     NEW_TASK = EVENTS_ENUM_NEW_TASK
     TASK_STATUS_UPDATE = EVENTS_ENUM_TASK_STATUS_UPDATE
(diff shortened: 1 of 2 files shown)


dbrasseur-aneo:
I'm fine with the changes, with 2 caveats:
- Why remove the "name from value" ? Displaying that the status is "2" is not all that helpful (and that's most likely what you'll have when constructing the task from the grpc message most likely). You may need to properly convert the int to the enum when constructing from a message, with perhaps an overload to __str__ on the enums to display it properly
- Not sure the ruff change belongs here

qdelamea-aneo:
> I'm fine with the changes, with 2 caveats:
> 
> * Why remove the "name from value" ? Displaying that the status is "2" is not all that helpful (and that's most likely what you'll have when constructing the task from the grpc message most likely). You may need to properly convert the int to the enum when constructing from a message, with perhaps an overload of `__str__` on the enums to display it properly
> * Not sure the ruff change belongs here

You're right!
* To avoid any breakage or other problems, I finally kept the name_from_value methods.
* For Ruff configuration, I created a separate pr (see (#572)[https://github.com/aneoconsulting/ArmoniK.Api/pull/572]).

dbrasseur-aneo (approved):

PR #583, opened 2025-01-09
feat: add tcp endpoint for grpc channel

jfonseca-aneo:
# Motivation

Java does not support grpc communication  between the worker and the agent over unix sockets, this PR adds the option to use a tcp socket instead. 

# Description

Enlarged the kestrel configuration options in the worker server so it also accepts to listen on a tcp socket. 
# Testing

Not applicable

# Impact

Carry on similar logic on the server configuration on the Agent in Armonik.Core

# Additional Information

Not applicable

# Checklist

### DIFF
--- a/packages/csharp/ArmoniK.Api.Common/Options/GrpcChannel.cs
+++ b/packages/csharp/ArmoniK.Api.Common/Options/GrpcChannel.cs
@@ -40,4 +40,9 @@ public class GrpcChannel
   ///   Type of gRPC Socket used
   /// </summary>
   public GrpcSocketType SocketType { get; set; } = GrpcSocketType.UnixDomainSocket;
+
+  /// <summary>
+  ///   Port for communicate with this gRPC Channel in case a tcp socket is used
+  /// </summary>
+  public int TcpEndPoint { get; set; }
 }
--- a/packages/csharp/ArmoniK.Api.Common/Options/GrpcChannel.cs
+++ b/packages/csharp/ArmoniK.Api.Common/Options/GrpcChannel.cs
@@ -40,9 +40,4 @@ public class GrpcChannel
   ///   Type of gRPC Socket used
   /// </summary>
   public GrpcSocketType SocketType { get; set; } = GrpcSocketType.UnixDomainSocket;
-
-  /// <summary>
-  ///   Port for communicate with this gRPC Channel in case a tcp socket is used
-  /// </summary>
-  public int TcpEndPoint { get; set; }
 }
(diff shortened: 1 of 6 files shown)


packages/csharp/ArmoniK.Api.Common/Options/GrpcChannel.cs
```diff
@@ -40,4 +40,9 @@ public class GrpcChannel
   ///   Type of gRPC Socket used
   /// </summary>
   public GrpcSocketType SocketType { get; set; } = GrpcSocketType.UnixDomainSocket;
+
+  /// <summary>
+  ///   Port for communicate with this gRPC Channel in case a tcp socket is used
```
aneojgurhem, line 45: `Address` should be used instead

jfonseca-aneo: If I understood correctly, even if we choose to do the communication via a tcp socket, the agent should still create the unix socket because its absence is used to decide whether to stop other (listening) processes. This is why I added a separate field to hold the tcp port.

aneojgurhem (commented):
You may want to do something similar in Agent Program.cs in Core. I'd also try to factorize the code with the GrpcChannelProvider

packages/csharp/ArmoniK.Api.Common/Options/GrpcChannel.cs
aneojgurhem: Oh yeah there is that. To bypass this issue, I'd transform worker container in Kubernetes POD into a sidecar container (the same as for fluentbit) as it would guarantee that the worker starts before the agent and stops after it.

aneojgurhem: As of today, the Kubernetes provider we use for terraform does not support to do so. However, with the direct use of Helm Charts, we will be able to do so.

ngruelaneo:
```suggestion
  ///   Port to communicate with this gRPC Channel in case a tcp socket is used
```

aneojgurhem (approved):

packages/csharp/ArmoniK.Api.Common/Options/GrpcChannel.cs
jfonseca-aneo: @ngruelaneo thanks for the suggestion, this code was actually not necessary.

PR #590, opened 2025-01-30
chore: parametrize keep alive options

jfonseca-aneo:
# Motivation

PR https://github.com/aneoconsulting/ArmoniK.Api/pull/583 added support to allow the usage of a tcp socket between the agent and the worker. There are still issues with connection (for Java grpc) which are solved deactivating some keep-alive options   . This PR exposes those options so they can be set via environmental variables during deployment. 

# Description

Added the two relevant Keep-alive options  as members of the `GrpcChannel` class, the options are deactivated by setting
them as `TimeSpan.MaxValue`. 

In order to allow the user to specify this in the corresponding environment variables, for example to disable the `KeepAlivePingTimeout` in the `WorkerChannel`,  we make use of the environment variable:

```hcl 

"ComputePlane__WorkerChannel__KeepAlivePingTimeout=MaxValue"

```
 
manually parse the `ComputePlane` configuration to account this and correctly parse the string "MaxValue" as `TimeSpan.MaxValue`  before registering the configuration in the dependency injection container. If the above mentioned
environment variable is not defined we take care to set a default value that matches the default value of the [documentation](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.core.http2limits.keepalivepingtimeout?view=aspnetcore-9.0&viewFallbackFrom=net-8.0) 

# Testing

A unit test has been added to verify that the configuration is properly read and values correctly parsed.

# Impact

In the case the user wants to have control over this options, it is necessary to define the following environment variables

```hcl
    "ComputePlane__WorkerChannel__KeepAlivePingTimeout=MaxValue",
    "ComputePlane__WorkerChannel__KeepAliveTimeout=MaxValue",
    "ComputePlane__AgentChannel__KeepAlivePingTimeout=00:01:00",
    "ComputePlane__AgentChannel__KeepAliveTimeout=00:00:20"
```

The chosen values are just an example, only a valid TimeSpan or the string "MaxValue" are accepted.
These are only used for tcp sockets. 

# Additional Information

None.

# Checklist

### DIFF
--- a/packages/csharp/ArmoniK.Api.Common/Options/ComputePlane.cs
+++ b/packages/csharp/ArmoniK.Api.Common/Options/ComputePlane.cs
@@ -41,12 +41,12 @@ public class ComputePlane
   /// <summary>
   ///   Path to the section containing the worker channel values in configuration object
   /// </summary>
-  public const string WorkerChannelSection = "WorkerChannel";
+  public const string WorkerChannelSection = nameof(WorkerChannel);
 
   /// <summary>
   ///  Path to the section containing the agent channel values in configuration object
   /// </summary>
-  public const string AgentChannelSection = "AgentChannel";
+  public const string AgentChannelSection = nameof(AgentChannel);
 
   /// <summary>
   ///   Channel used by the Agent to send tasks to the Worker
--- a/packages/csharp/ArmoniK.Api.Common/Options/ComputePlane.cs
+++ b/packages/csharp/ArmoniK.Api.Common/Options/ComputePlane.cs
@@ -44,7 +44,7 @@ public class ComputePlane
   public const string WorkerChannelSection = nameof(WorkerChannel);
 
   /// <summary>
-  ///  Path to the section containing the agent channel values in configuration object
+  ///   Path to the section containing the agent channel values in configuration object
   /// </summary>
   public const string AgentChannelSection = nameof(AgentChannel);
 
--- a/packages/csharp/ArmoniK.Api.Common/Options/ComputePlane.cs
+++ b/packages/csharp/ArmoniK.Api.Common/Options/ComputePlane.cs
@@ -39,6 +39,16 @@ public class ComputePlane
   public const string SettingSection = nameof(ComputePlane);
 
   /// <summary>
+  ///   Path to the section containing the worker channel values in configuration object
+  /// </summary>
+  public const string WorkerChannelSection = "WorkerChannel";
+
+  /// <summary>
+  ///  Path to the section containing the agent channel values in configuration object
+  /// </summary>
+  public const string AgentChannelSection = "AgentChannel";
+
+  /// <summary>
   ///   Channel used by the Agent to send tasks to the Worker
   /// </summary>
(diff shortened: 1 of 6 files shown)


packages/csharp/ArmoniK.Api.Common/Options/ComputePlane.cs
```diff
@@ -38,6 +38,16 @@ public class ComputePlane
...
 
+  /// <summary>
+  ///   Path to the section containing the worker channel values in configuration object
+  /// </summary>
+  public const string WorkerChannelSection = "WorkerChannel";
+
+  /// <summary>
+  ///  Path to the section containing the agent channel values in configuration object
+  /// </summary>
+  public const string AgentChannelSection = "AgentChannel";
```
aneojgurhem, line 49: Is it possible to put a `nameof` here ?
```suggestion
  public const string AgentChannelSection = nameof(AgentChannel);
```

packages/csharp/ArmoniK.Api.Worker/Utils/WorkerServer.cs
```diff
@@ -133,10 +133,43 @@ public static WebApplication Create<T>(Action<IServiceCollection, IConfiguration
...
+                                               }
+
+                                               if (TimeSpan.TryParse(toParse,
+                                                                     out var result))
+                                               {
+                                                 return result;
+                                               }
+
+                                               throw new FormatException($"Provide a valid TimeSpan format: {toParse} or 'MaxValue'");
+                                             }
```
aneojgurhem, line 172: Dependency Injection already does that for you

aneojgurhem: see https://github.com/aneoconsulting/ArmoniK.Core/blob/main/Utils/src/ConfigurationExt.cs

jfonseca-aneo: Ok, but this does not allow to handle the conversion from "MaxValue" to `TimeSpan.MaxValue` which is the correct value to put to desactivate a given keep alive option, and neither setting a default value. Another point is that the keep alive options might not be processed by the DI at this point, at least is not clear to me ^^

packages/csharp/ArmoniK.Api.Common/Utils/ConfigurationExt.cs
```diff
@@ -0,0 +1,94 @@
...
+  /// <typeparam name="T">Type of the options class</typeparam>
+  /// <param name="configuration">Configurations used to populate the class</param>
+  /// <param name="key">Path to the Object in the configuration</param>
+  /// <returns>
+  ///   The initialized object
+  /// </returns>
+  public static T GetInitializedValue<T>(this IConfiguration configuration,
+                                         string              key)
+    where T : new()
+    => configuration.GetSection(key)
```
aneojgurhem, line 57: you do not seem to need those 2.

aneojgurhem (approved):

PR #597, opened 2025-02-21
fix: add missing method to wait for result completion

aneojgurhem:
# Motivation

[Include the reason behind these changes and any relevant context.]

# Description

[Provide a detailled explanation of the modifications you have made. Link any related issues.]

# Testing

[When applicable, detail the testing you have performed to ensure that these changes function as intended. Include information about any added tests.]

# Impact

[Discuss the impact of your modifications on ArmoniK. This might include effects on performance, configuration, documentation, new dependencies, or changes in behaviour.]

# Additional Information

[Any additional information that reviewers should be aware of.]

# Checklist


packages/java/src/main/java/armonik/client/event/EventClient.java
```diff
@@ -182,4 +75,99 @@ public static EventSubscriptionRequest CreateEventSubscriptionRequest(String ses
...
+      int bucketSize,
+      int parallelism) {
+    List<List<String>> chunks = chunkList(new ArrayList<>(resultIds), bucketSize);
+    ExecutorService executor = Executors.newFixedThreadPool(parallelism);
+    List<CompletableFuture<Void>> futures = new ArrayList<>();
+
+    for (List<String> chunk : chunks) {
+      futures.add(CompletableFuture.runAsync(() -> {
+        Set<String> resultsNotFound = new HashSet<>(chunk);
+        while (!resultsNotFound.isEmpty()) {
```
jdongo-aneo, line 108: There is a risk of infinite loop if no matching events are received; We can use a timeout, but what should be the duration 
? As we can have tasks that take very long time to complete;

aneojgurhem: Yes we can wait indefinitely if session is not cancelled by the gui for instance. That is intentional as, usually, we do not know that there are issues client side.
These methods are helpers for common actions. They can be rewritten if more specific needs arise when integrating an application into ArmoniK

PR #607, opened 2025-03-21
test: Added Worker calls to Api.Mock

dbrasseur-aneo:
# Motivation

Some tests needed an agent to call a worker. 

# Description

This gives the Api.Mock this capability, and python received an example test for this new test

# Testing

It's a test

# Impact

No impact, it's tests only

# Additional Information

# Checklist


lemaitre-aneo (commented):
In the pipeline, I would split client tests from worker tests as I do not expect workers to use tls at all, and I expect other languages do not support tls on the worker side currently.

PR #609, opened 2025-03-26
feat: Python Worker can now connect to agent using certificates

dbrasseur-aneo:
# Motivation

Agent and worker creation was inconsistent with the rest of endpoints declarations, espcially using http.

# Description

Standardises the endpoint parsing with the previous parser and extends the change to allow for a orker to communicate with the agent using certificates

# Testing

Testing is done in a specific other PR that has other implications

# Impact

No impact to existing code. New code can add certificates.

# Additional Information

[Any additional information that reviewers should be aware of.]

# Checklist


packages/python/tests/conftest.py
```diff
@@ -15,9 +15,9 @@
 from armonik.protogen.worker.agent_service_pb2_grpc import AgentStub
 from typing import List, Union
 
-ca_cert = os.getenv("Grpc__CaCert")
-client_cert = os.getenv("Grpc__ClientCert")
-client_key = os.getenv("Grpc__ClientKey")
+ca_cert = os.getenv("Grpc__CaCert") or None
```
lemaitre-aneo, line 18: Why don't you check GrpcClient__CaCert?
What is the purpose of `or None`?

packages/python/src/armonik/common/channel.py
```diff
@@ -76,6 +76,33 @@ def _load_certificates(
...
+
+    if "unix" in scheme:
+        # gRPC supports unix:path and the path is then relative, if the scheme is unix://, then the path is absolute
+        if endpoint.startswith("/"):
+            endpoint = "unix://" + endpoint
+        else:
+            endpoint = "unix:" + endpoint
+    elif keep_scheme:
+        endpoint = parsed.scheme + "://" + endpoint
+    return scheme, endpoint
```
lemaitre-aneo, line 103: I'm not sure to follow the logic. Why do you need this?

packages/python/tests/conftest.py
dbrasseur-aneo: Forgot the Client_CaCert
The "or None" is to force None if the environement variable is set but empty

packages/python/src/armonik/common/channel.py
dbrasseur-aneo: So it can be used elsewhere (notably in the worker creation) because depending on the context, the expected endpoint is either the netloc+path, or scheme+netloc+path

PR #631, opened 2025-08-19
feat: Added get_resource_data implementation

AncientPatata:
# Motivation

It wasn't implemented prior to this, but is clearly needed in projects like PymoniK wherein it was implemented directly.

# Description

- Added implementation of get_resource_data.

# Testing

Previous tests are passing.

# Impact

Not applicable.

# Additional Information

Not applicable.

# Checklist


packages/python/src/armonik/worker/taskhandler.py
```diff
@@ -194,3 +197,24 @@ def create_results(
             for message in response.results:
                 results[message.name] = Result.from_result_metadata(message)
         return results
+
+    def get_resource_data(self, result_id: str) -> Path:
```
lemaitre-aneo, line 201: Why the function returns a Path instead of `bytes`?

AncientPatata: I don't know what the spec is on GetResourcePath. When I checked in Core, GetResourceData just downloads the file to your data folder and then it's up to the user to read this file. In PymoniK, because it's an abstraction over the API, I do provide an option for returning the bytes directly; that makes a lot of sense. But it was also not the only option because, even from the onset (materializing files / caching files), having a file path is pretty useful. 

Afterall, If the user wanted to upload a file (example of a zip file) then it'd make sense to end up with a file there instead of reading the bytes and then having the user rewrite the same bytes to disk. They can choose to move the file if they choose so.  

TL;DR: With no spec of how GetResourceData should be implemented in the APIs; this seemed to be the most generic way of implementing it. There are many different ways of going about this (one or the other / hybrid approach) though and I'd like to hear your thoughts.

PR #650, opened 2025-12-02
feat: Draft: Session status notification stream

plpn90:
// DRAFT: also did some side modifications which i can undo before merge

# Motivation

Needed a better way to observe changes to sessions being opened/modified/deleted, other than polling the` ListSessions()` endpoint on the` SessionsService`.

# Description

Adds another endpoint on the `EventsService` to create a notification channel for session updates.

# Impact

Similiar idea to how the `ChangeStream` is implemented for `Results`. Btw, maybe it is time to rethink how these notifications work. Wouldn't MessageQueues be a better alternative for notifying changes/updates to the system overall? Also makes it independent to MongoDB (and for it being mandatory to be a replicaset!)

### DIFF
--- a/Protos/V1/events_common.proto
+++ b/Protos/V1/events_common.proto
@@ -6,6 +6,7 @@ import "result_status.proto";
 import "results_filters.proto";
 import "task_status.proto";
 import "tasks_filters.proto";
+import "session_status.proto";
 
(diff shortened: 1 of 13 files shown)


aneojgurhem (changes requested):
Please remove changes unrelated to this PR

PR #671, opened 2026-05-18
feat: Add configurable gRPC retry policies to worker channels

italo1aguiar-cacib:
# Motivation

There is no retry policy on grpc connection in place, this causes problems. Other changes will be made on ArmoniK Core to better manage the connection between agent - worker. 

# Description

Add configurable gRPC retry policies to worker channels.

# Testing

Unit tests and test

# Impact

Makes worker - agent connection more resilient 

# Additional Information

# Checklist

### DIFF
--- a/packages/csharp/ArmoniK.Api.Common.Channel/Utils/GrpcChannelProvider.cs
+++ b/packages/csharp/ArmoniK.Api.Common.Channel/Utils/GrpcChannelProvider.cs
@@ -1,13 +1,13 @@
 // This file is part of the ArmoniK project
-// 
+//
 // Copyright (C) ANEO, 2021-2026. All rights reserved.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License")
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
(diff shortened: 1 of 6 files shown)


aneojgurhem:
Thanks for the work on this. The motivation is valid — the Worker/HealthCheck call failing before the worker is ready is a real problem. However, the fix is applied in the wrong place.

Worker/HealthCheck is called by the agent on the worker. The existing retry mechanic around that call in ArmoniK.Core is the right place to address it.

The current approach applies MethodName.Default on the AgentChannel (worker→agent), which retries all worker→agent RPCs including non-idempotent ones: SubmitTasks, CreateResults, CreateResultsMetaData, NotifyResultData. Retrying those on a transient failure where the agent already processed the request will create duplicate tasks or results.

On the ArmoniK.Core side, a channel-level retry would also be the wrong approach: MethodName.Default would catch Worker/Process in addition to Worker/HealthCheck. ArmoniK.Core creates one agent instance per task execution, so retrying Worker/Process means the task body runs multiple times within the same agent. Any SubmitTasks, CreateResults, or NotifyResultData calls the task makes would be issued again by that same agent, producing duplicates that ArmoniK's task-level retry logic is not designed to handle.

I'd suggest closing this PR and instead tuning the retry behavior in ArmoniK.Core's existing Worker/HealthCheck retry logic.

PR #673, opened 2026-05-27
fix(python): replace removed setuptools_scm private import with packaging.version

aneojgurhem:
## Motivation

`setuptools_scm._version_cls._version_as_tuple` is a private internal function that was removed in newer versions of `setuptools_scm`, causing the Python package build pipeline to fail with:

```
ModuleNotFoundError: No module named 'setuptools_scm._version_cls'
```

## Description

- Removes the broken `from setuptools_scm._version_cls import _version_as_tuple` import.
- Adds `from packaging.version import Version as PkgVersion` — `packaging` is a mandatory transitive dependency of `setuptools_scm` and is always available.
- Reimplements `_version_as_tuple(version_str)` as a local function using `packaging.version.Version`, deriving the tuple directly from the version string returned by `_get_version` (rather than as a side-effect of the version-scheme callback, which could silently produce an empty tuple if the callback was never invoked).
- Fixes import ordering (`argparse` moved to the top-level imports block).

## Testing

No automated tests exist for this script. The fix can be verified by running:

```sh
python genversion.py --write-to /tmp/test_version.py
cat /tmp/test_version.py
```

and confirming that both `__version__` and `__version_tuple__` contain consistent, non-empty values.

## Impact

- No behaviour change for the normal build path.
- Removes dependency on a private `setuptools_scm` internal; the script now only relies on `packaging`, which is stable public API.

## Additional Information

The reimplemented `_version_as_tuple` handles release segments (as `int`) and dev segments (as `"devN"` string), matching the convention used by the original function.


packages/python/genversion.py
```diff
@@ -1,9 +1,9 @@
 import sys
+from argparse import ArgumentParser
 
 from setuptools_scm.version import ScmVersion, guess_next_version
 from setuptools_scm import Configuration, _get_version
```
lemaitre-aneo: We will most likely have the problem with this one in the future, don't we?

packages/python/genversion.py
```diff
@@ -23,6 +23,14 @@ def get_version(self, version: ScmVersion):
         return version.format_with("{tag}")
 
 
+def _version_as_tuple(version_str: str) -> tuple:
```
lemaitre-aneo: If we implement it ourselves, I think it should not be prefixed with an underscore.

packages/python/genversion.py
aneojgurhem: yes

lemaitre-aneo (approved):

PR #680, opened 2026-06-25
feat: self contained implementation of string view

jfonseca-aneo:
# Motivation

  ArmoniK's Cpp API currently depends on Abseil (`absl::string_view`, `absl::StrSplit`, `absl::StrContains`) for string handling in its own code. Abseil is already a transitive dependency of protobuf/gRPC and cannot be removed from the build graph, but it should not appear in ArmoniK's own headers and sources. Using `absl::string_view` in public headers  forces every downstream consumer to have Abseil available and visible, ties the ABI to  Abseil's versioning, and prevents C++11-clean compilation without a C++17 standard library.

  # Description

  Introduces a self-contained, header-only `armonik::api::common::string_view` type (`ArmoniK.Api.Common/header/utils/string_view.h`) and replaces every Abseil string-utility call across the C++ package with it.

  # Testing

  - New `StringViewTest.cpp` added with comprehensive unit tests covering all methods and edge
  cases.
  - Existing test suite (`LoggerTest`, `MockTest`, `ResultsClientTest`) updated to use 
  `armonik::api::string_view`; all tests continue to compile and pass against the new type.

  # Impact

  - **No behaviour change**: `armonik::api::common::string_view` is ABI-compatible with  `absl::string_view` at call sites — same two-word layout, same semantics.
  - **No CMake changes**: Abseil link targets are preserved; only `#include` directives in  ArmoniK's own files change.
  - **Downstream headers**: Public headers no longer expose `absl::` identifiers, removing the implicit requirement for consumers to have Abseil on their include path.
  - **C++ standard**: code remains C++14 (`CXX_STANDARD` unchanged); the new type is C++11-compatible.
  - **New header**: `ArmoniK.Api.Common/header/utils/string_view.h` — part of the installed Common interface.

  # Additional Information

  The implementation follows the standard C++ idioms; the non-obvious (or rather old) techniques carry inline cppreference links:
  - `static_cast<size_type>(-1)` for `npos` — same as the standard definition.
  - Recursive `clen` helper for `constexpr strlen` — required by C++11's single-return 
  constraint.
  - Nested ternary in `level_name()` — same constraint applied to the switch-based logger 
  level converter.

  `absl` still appears in the build graph as a transitive dependency of gRPC/protobuf and in  generated `.pb.` files.
  # Checklist

### DIFF
--- a/packages/cpp/ArmoniK.Api.Client/source/results/ResultsClient.cpp
+++ b/packages/cpp/ArmoniK.Api.Client/source/results/ResultsClient.cpp
@@ -6,9 +6,9 @@ namespace armonik {
 namespace api {
 namespace client {
 
-std::map<std::string, std::string> ResultsClient::create_results(absl::string_view session_id,
+std::map<std::string, std::string> ResultsClient::create_results(common::string_view session_id,
                                                                  const std::vector<std::string> &names) {
-  return create_results_metadata(std::string(session_id), names);
+  return create_results_metadata(std::string(session_id.data(), session_id.size()), names);
 }
 
 std::map<std::string, std::string> ResultsClient::create_results_metadata(std::string session_id,
@@ -41,7 +41,7 @@ std::map<std::string, std::string> ResultsClient::create_results_metadata(std::s
   return mapping;
 }
 
-void ResultsClient::upload_result_data(std::string session_id, std::string result_id, absl::string_view payload) {
+void ResultsClient::upload_result_data(std::string session_id, std::string result_id, common::string_view payload) {
 
   size_t maxChunkSize = get_service_configuration().data_chunk_max_size;
 
--- a/packages/cpp/ArmoniK.Api.Client/source/results/ResultsClient.cpp
+++ b/packages/cpp/ArmoniK.Api.Client/source/results/ResultsClient.cpp
@@ -8,7 +8,7 @@ namespace client {
 
 std::map<std::string, std::string> ResultsClient::create_results(common::string_view session_id,
                                                                  const std::vector<std::string> &names) {
-  return create_results_metadata(std::string(session_id.data(), session_id.size()), names);
+  return create_results_metadata(static_cast<std::string>(session_id), names);
 }
 
 std::map<std::string, std::string> ResultsClient::create_results_metadata(std::string session_id,
--- a/packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
+++ b/packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
@@ -1,0 +1,152 @@
+#pragma once
+/**
+ * @file string_view.h
+ * @brief A C++11-compatible non-owning string view type.
+ */
+
+#include <cstddef>
+#include <ostream>
+#include <string>
+
+namespace armonik {
+namespace api {
+namespace common {
+
+/**
+ * @brief Non-owning reference to a string.
+ *
+ * ABI-stable across C++ standards: always the same concrete type regardless of
+ * whether the translation unit is compiled as C++11, 14, or 17. Optional
+ * conversions to/from std::string_view are guarded by the feature macro so they
+ * are compiled only on C++17 targets.
+ */
+class string_view {
+public:
+  using const_iterator = const char *;
+  using size_type = std::size_t;
+
+  // Casting -1 to an unsigned type yields its maximum representable value. See: 
+  // https://en.cppreference.com/w/cpp/string/basic_string_view/npos
+  static constexpr size_type npos = static_cast<size_type>(-1);
+
+private:
+  const char *data_;
+  size_type size_;
+
+  // C++11 constexpr functions are restricted to a single return statement, so a
+  // loop-based strlen is not valid: https://en.cppreference.com/w/cpp/language/constexpr (see "C++11" notes section)
+  static constexpr size_type clen(const char *s) noexcept { return *s ? 1 + clen(s + 1) : 0; }
+
+public:
+  constexpr string_view() noexcept : data_(nullptr), size_(0) {}
+
+  constexpr string_view(const char *s) noexcept : data_(s), size_(clen(s)) {}
+
+  constexpr string_view(const char *s, size_type len) noexcept : data_(s), size_(len) {}
+
+  string_view(const std::string &s) noexcept : data_(s.data()), size_(s.size()) {}
+
+  constexpr string_view(const string_view &) noexcept = default;
+  string_view &operator=(const string_view &) noexcept = default;
+
+  // --- Accessors ---
+  constexpr const char *data() const noexcept { return data_; }
+  constexpr size_type size() const noexcept { return size_; }
+  constexpr bool empty() const noexcept { return size_ == 0; }
+  constexpr const char &operator[](size_type i) const noexcept { return data_[i]; }
+
+  // --- Iterators ---
+  constexpr const_iterator begin() const noexcept { return data_; }
+  constexpr const_iterator end() const noexcept { return data_ + size_; }
+  constexpr const_iterator cbegin() const noexcept { return data_; }
+  constexpr const_iterator cend() const noexcept { return data_ + size_; }
+
+  // --- Substr ---
+  // Single return for C++11 constexpr compatibility
+  constexpr string_view substr(size_type pos, size_type len = npos) const noexcept {
+    return string_view(data_ + pos, (len == npos || pos + len > size_) ? size_ - pos : len);
+  }
+
+  // --- Find ---
+  size_type find(char c, size_type pos = 0) const noexcept {
+    for (size_type i = pos; i < size_; ++i) {
+      if (data_[i] == c) {
+        return i;
+      }
+    }
+    return npos;
+  }
+
+  size_type find(const string_view &needle, size_type pos = 0) const noexcept {
+    if (pos > size_) {
+      return npos;
+    }
+    if (needle.size_ == 0) {
+      return pos;
+    }
+    if (needle.size_ > size_ - pos) {
+      return npos;
+    }
+    for (size_type i = pos; i + needle.size_ <= size_; ++i) {
+      bool match = true;
+      for (size_type j = 0; j < needle.size_; ++j) {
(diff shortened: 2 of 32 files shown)


packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
```diff
@@ -0,0 +1,154 @@
...
+ * are compiled only on C++17 targets.
+ */
+class string_view {
+public:
+  using const_iterator = const char *;
+  using size_type = std::size_t;
+
+  // Casting -1 to an unsigned type yields its maximum representable value. See:
+  // https://en.cppreference.com/w/cpp/string/basic_string_view/npos
+  static constexpr size_type npos = static_cast<size_type>(-1);
```
lemaitre-aneo: Prefer using std::string::npos

packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
```diff
@@ -0,0 +1,154 @@
...
+
+private:
+  const char *data_;
+  size_type size_;
+
+  // C++11 constexpr functions are restricted to a single return statement, so a
+  // loop-based strlen is not valid: https://en.cppreference.com/w/cpp/language/constexpr (see "C++11" notes section)
+  static constexpr size_type clen(const char *s) noexcept { return *s ? 1 + clen(s + 1) : 0; }
```
lemaitre-aneo, line 45: Do we really use clen?

packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
```diff
@@ -0,0 +1,154 @@
...
+  constexpr string_view() noexcept : data_(nullptr), size_(0) {}
+
+  constexpr string_view(const char *s) noexcept : data_(s), size_(clen(s)) {}
+
+  constexpr string_view(const char *s, size_type len) noexcept : data_(s), size_(len) {}
+
+  string_view(const std::string &s) noexcept : data_(s.data()), size_(s.size()) {}
+
+  constexpr string_view(const string_view &) noexcept = default;
+  string_view &operator=(const string_view &) noexcept = default;
```
lemaitre-aneo, line 57: There is a missing constructor from "string literal"

packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
```diff
@@ -0,0 +1,154 @@
...
+  // --- Iterators ---
+  constexpr const_iterator begin() const noexcept { return data_; }
+  constexpr const_iterator end() const noexcept { return data_ + size_; }
+  constexpr const_iterator cbegin() const noexcept { return data_; }
+  constexpr const_iterator cend() const noexcept { return data_ + size_; }
+
+  // --- Substr ---
+  // Single return for C++11 constexpr compatibility
+  constexpr string_view substr(size_type pos, size_type len = npos) const noexcept {
+    return string_view(data_ + pos, (len == npos || pos + len > size_) ? size_ - pos : len);
```
lemaitre-aneo: What if pos is larger than size?

packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
```diff
@@ -0,0 +1,154 @@
...
+// --- Free operators ---
+inline bool operator==(const char *lhs, const string_view &rhs) noexcept { return rhs == lhs; }
+inline bool operator!=(const char *lhs, const string_view &rhs) noexcept { return rhs != lhs; }
+
+inline std::ostream &operator<<(std::ostream &os, const string_view &sv) {
+  if (!sv.empty()) {
+    os.write(sv.data(), static_cast<std::streamsize>(sv.size()));
+  }
+  return os;
+}
```
lemaitre-aneo: I would go with friend functions

packages/cpp/ArmoniK.Api.Client/source/results/ResultsClient.cpp
```diff
@@ -6,9 +6,9 @@ namespace armonik {
 namespace api {
 namespace client {
 
-std::map<std::string, std::string> ResultsClient::create_results(absl::string_view session_id,
+std::map<std::string, std::string> ResultsClient::create_results(common::string_view session_id,
                                                                  const std::vector<std::string> &names) {
-  return create_results_metadata(std::string(session_id), names);
+  return create_results_metadata(std::string(session_id.data(), session_id.size()), names);
```
lemaitre-aneo: Use static_cast

packages/cpp/ArmoniK.Api.Common/source/utils/Utils.cpp
```diff
@@ -1,8 +1,23 @@
...
+  std::vector<std::string> result;
+  std::string::size_type start = 0, pos;
+  while ((pos = s.find(delim, start)) != std::string::npos) {
+    result.emplace_back(s.begin() + start, s.begin() + pos);
+    start = pos + 1;
+  }
+  result.emplace_back(s.begin() + start, s.end());
+  return result;
+}
+static bool str_contains(const std::string &s, char c) { return s.find(c) != std::string::npos; }
```
lemaitre-aneo: I would put this within string_view

packages/cpp/ArmoniK.Api.Common/source/utils/Utils.cpp
```diff
@@ -1,8 +1,23 @@
 #include "utils/Utils.h"
-#include <absl/strings/str_split.h>
 #include <cmath>
 #include <iomanip>
 
+/// @cond DOXYGEN_IGNORE
+namespace {
+static std::vector<std::string> str_split(const std::string &s, char delim) {
```
lemaitre-aneo: This could be implemented for string_view

packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
jfonseca-aneo: Not in the rest of the API or in the SDK, nevertheless I'd keep it to protect our selves for the case of a plain pointer coming from pure C code.

packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
```diff
@@ -0,0 +1,194 @@
...
+  size_type find(char c, size_type pos = 0) const noexcept {
+    for (size_type i = pos; i < size_; ++i) {
+      if (data_[i] == c) {
+        return i;
+      }
+    }
+    return npos;
+  }
+
+  size_type find(const string_view &needle, size_type pos = 0) const noexcept {
```
lemaitre-aneo:
```suggestion
  size_type find(string_view needle, size_type pos = 0) const noexcept {
```

packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
```diff
@@ -0,0 +1,194 @@
...
+      }
+    }
+    return npos;
+  }
+
+  size_type find(const char *s, size_type pos = 0) const noexcept { return find(string_view(s), pos); }
+
+  template <std::size_t N> size_type find(const char (&s)[N], size_type pos = 0) const noexcept {
+    return find(string_view(s, N - 1), pos);
+  }
```
lemaitre-aneo: No need for these overloads if the base implementation takes string_view by value

packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
```diff
@@ -0,0 +1,194 @@
...
+    while ((pos = find(delim, start)) != npos) {
+      result.emplace_back(data_ + start, pos - start);
+      start = pos + 1;
+    }
+    result.emplace_back(data_ + start, size_ - start);
+    return result;
+  }
+
+  // --- Comparisons ---
+  bool operator==(const string_view &other) const noexcept {
```
lemaitre-aneo: If you define it friend with pass-by-value semantics, you only have to define it once, and all conversions will be handled without issues:

```suggestion
  friend bool operator==(string_view lhs, string_view rhs) noexcept {
```

packages/cpp/ArmoniK.Api.Common/header/utils/string_view.h
```diff
@@ -0,0 +1,194 @@
...
+  // C++11 constexpr functions are restricted to a single return statement, so a
+  // loop-based strlen is not valid: https://en.cppreference.com/w/cpp/language/constexpr (see "C++11" notes section)
+  static constexpr size_type clen(const char *s) noexcept { return *s ? 1 + clen(s + 1) : 0; }
+
+public:
+  constexpr string_view() noexcept : data_(nullptr), size_(0) {}
+
+  constexpr string_view(const char *s) noexcept : data_(s), size_(clen(s)) {}
+
+  template <std::size_t N> constexpr string_view(const char (&s)[N]) noexcept : data_(s), size_(N - 1) {}
```
lemaitre-aneo: Apparently, this does not do what I recalled from memory, and has no precedence over the `const char*` overload. We should discuss what we really want here.

packages/cpp/ArmoniK.Api.Common/source/utils/Utils.cpp
```diff
@@ -22,41 +22,40 @@ ::google::protobuf::Duration duration_from_values(long long int days, long long
...
     throw std::invalid_argument("timespan is not of the format [-][d.]hh:mm:ss[.fffffffff]");
   }
   // Sign is only present in the first section
-  int sign = absl::StrContains(subsplit[0], '-') ? -1 : 1;
+  int sign = subsplit[0].contains('-') ? -1 : 1;
   if (subsplit.size() == 2) {
-    days = std::strtol(subsplit[0].c_str(), nullptr, 10);
-    hours = sign * std::strtol(subsplit[1].c_str(), nullptr, 10);
+    days = std::strtol(static_cast<std::string>(subsplit[0]).c_str(), nullptr, 10);
+    hours = sign * std::strtol(static_cast<std::string>(subsplit[1]).c_str(), nullptr, 10);
```
lemaitre-aneo: Ouch, I did not anticipate that...

lemaitre-aneo (approved):

PR #689, opened 2026-07-30
refactor(rust): make packages/rust a workspace and extract armonik-transport

wkirschenmann:
﻿# Motivation

`armonik` mixes two concerns. One is getting a configured, TLS-terminated channel to an ArmoniK
endpoint: parsing `GrpcClient__*`, loading certificates, building the `rustls` config, applying keepalive
and timeout settings. The other is the generated types and services that talk over that channel.

Anyone who wants only the first has to take the second, and the second means `protoc` and a
`tonic-prost-build` step compiling forty `.proto` files they will never call.

Splitting them also puts a name on something that was already true: the connection layer has no
dependency on anything ArmoniK-specific. It is a `ClientConfig` and a `tonic::transport::Channel`, and it
was only ever in this crate because that is where it was first written.

# Description

`armonik-transport` is the first concern, moved out. `armonik` depends on it and re-exports every type at
the path it already had — `armonik::ClientConfig`, `armonik::client::ClientConfigArgs`,
`armonik::client::ConfigError`, `armonik::client::ConnectionError`, `armonik::client::ReadEnvError` — so
nothing downstream has to know the split happened.

What moved, and only what moved:

| from | to |
|---|---|
| `armonik/src/client/config.rs` | `armonik-transport/src/config.rs` |
| the channel building in `Client::with_config`, and the private `Client::https_connector` under it | `connect()` and `https_connector()` in `armonik-transport/src/connect.rs` |
| `ConnectionError` | with them, since it describes exactly those failures |
| `read_env`, `read_env_bool`, `ReadEnvError`, `InsecureCertVerifier` from `utils.rs` | `armonik-transport/src/utils.rs` |

`config.rs` shows in the diff as a rename with a two-line change: one `super::ConfigError` becomes
`ConfigError`, now that it is in the module that defines it. `IntoCollection` and the serde helpers stay
in `armonik` — nothing in the transport layer uses them.

`Client::with_config` keeps its `tracing::debug_span!("Client", endpoint)` rather than the span moving
with the code it wrapped. The span is named after the type, and `armonik-transport` has no `Client`;
moving it would have renamed a span that appears in users' traces, for no reason beyond following the
code it happened to enclose.

Three consequences, named here rather than left to be discovered in review:

- **`https_connector` is now `pub`, and `#[doc(hidden)]`.** It was a private associated function.
  `armonik`'s `get_nb_request` test helper drives a raw HTTP request through the very same TLS stack a
  channel would use, in order to read the mock server's `/calls.json` — with the connector in another
  crate that is no longer reachable any other way. Hidden from the docs because its return type names
  `hyper-rustls` and `hyper-util` types rather than this crate's own.
- **`impl From<ConfigError> for ConnectionError` is new.** `Client::new` built that conversion with
  snafu's generated context selector, which does not cross a crate boundary.
- **`hyper` and `hyper-util` become dev-dependencies of `armonik`.** Its two remaining uses of
  `hyper::http::Request` are the same `http` crate reached through `tonic`, which already guarantees a
  matching version; `hyper` itself is now needed only by that test helper. `rustls`, `hyper-rustls` and
  `humantime` leave `armonik` altogether — they went with the code that used them.

## The workspace

The split makes this necessary rather than merely tidy: two crates that hand `rustls` and `tonic` types
to each other must not be able to resolve to two different copies of either.
`[workspace.dependencies]` pins those versions once, `[workspace.package]` holds the shared metadata and
the MSRV, and the lockfile moves to `packages/rust/Cargo.lock`, where a workspace's lockfile belongs.

## Two things that would otherwise have broken silently

**Version synchronisation.** `scripts/versions` read the Rust version from
`packages/rust/armonik/Cargo.toml`, which no longer carries one — it is `version.workspace = true`.
Unfixed, `update-versions` would have skipped Rust without saying so, and `verify-versions` would have
compared one fewer version, so a release could ship a stale crate version while looking clean. It now
reads `[workspace.package]`, and separately the `armonik-transport` dependency version, which has to
track the release for `cargo publish` to work at all.

**CI coverage.** `cargo test`/`build`/`clippy` ran from `packages/rust/armonik`, which covers one member.
Both workflows now work from `packages/rust` with `--workspace`; without that, the new crate would have
had no CI at all.

# Testing

The CI commands, as this PR changes them, run from `packages/rust`:

```
cargo build --workspace --locked                                              # ok
cargo fmt --all --check                                                       # clean
RUSTDOCFLAGS="-Dwarnings" cargo doc --workspace --no-deps                     # clean
cargo clippy --workspace --all-features --no-deps \
    -- -Dwarnings -Dunused-crate-dependencies                                 # clean
```

That last one is what caught `humantime` and `hyper-util` still being declared by `armonik` after their
users moved out.

**No test changes, and no behavioural claim beyond that.** `armonik`'s 97 library tests need the mock
server, so locally they fail on `GrpcClient__Endpoint` being unset — before and after this change alike:

```
before (main):  0 passed; 97 failed
after  (this branch):  0 passed; 97 failed
```

Same count, same reason, `InvalidUri(Empty)`. CI runs them properly through `scripts/mock_test.sh`, which
is where the real evidence for this PR is.

# Impact

- **No API change.** Every type is re-exported at its original path. A downstream crate compiles
  untouched; `armonik::reexports::rustls` still resolves, now routed through the transport crate so that
  it cannot be a different `rustls` from the one the connection was built with.
- **No behaviour change.** The connection code moved verbatim — same order, same settings, same errors,
  same tracing span.
- **`armonik` gets three fewer direct dependencies** (`rustls`, `hyper-rustls`, `humantime`), and two
  become dev-only.
- **Publishing order changes**: `armonik-transport` first, then `armonik`. A path dependency cannot be
  published, so `cargo publish` rewrites it into the version requirement in `armonik`'s manifest, and that
  version has to already be in the registry. Written down in the new crate's README.

# Additional Information

Two of the moved files read easily and one does not, which is worth knowing before starting.

`config.rs` is detected as a rename with a two-line diff, and `utils.rs` splits cleanly. `connect.rs`
does not: git sees 191 added lines, because the file it came from still exists — `client/mod.rs` lost 180
of its 555 lines rather than disappearing. To read it as the move it is:

```sh
git diff -M20% main...HEAD -- packages/rust        # lower the similarity threshold
git show main:packages/rust/armonik/src/client/mod.rs | sed -n '82,201p'   # the original, side by side
```

Line 82-121 of that file became `connect`, 127-201 became `https_connector`, and 476-513 is
`ConnectionError`. The three edits of substance are: the tracing span and the `Self::with_channel` wrapper
stayed behind in `Client::with_config` (the span is named after the type, and the wrapper builds a type
the transport crate does not have); `https_connector` went from private associated function to `pub`
function, having never used `Self`; and `crate::utils::InsecureCertVerifier` now resolves in the transport
crate, same text, different module.

The new crate has no tests of its own. The code it contains had none before the move either, and adding
them here would have made an extraction into something else. It is worth doing.

# Checklist

- [x] I have made corresponding changes to the documentation. *(a README for the new crate, covering what
      it is for and the publishing order; module docs on each moved file saying what it holds and why)*
- [ ] I have thoroughly tested my modifications and added tests when necessary. *(no test added — see
      Testing; this is a move, and the existing suite is the check that it is faithful)*
- [x] I have assessed the performance impact of my modifications. *(none: the same code runs in the same
      order. One more crate to compile, which `cargo` builds in parallel with `armonik`'s codegen.)*


wkirschenmann:
Need #687 and a rebase before merge to allow the CI to become green

packages/rust/armonik-transport/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+  "http2",
+  "native-tokio",
+  "tls12",
+  "logging",
+] }
+rustls = { workspace = true, features = ["ring", "logging", "std", "tls12"] }
+# `std` as well as `derive`: `default-features = false` drops it, and without it `String` implements
+# neither `Serialize` nor `Deserialize` — so the `serde` feature would not compile at all. It never did;
+# nothing built it, because CI's clippy pass did not enable any features.
+serde = { version = "1.0", features = [
```
lemaitre-aneo: Should most likely be a workspace dependency

packages/rust/armonik-transport/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+pub use config::{ClientConfig, ClientConfigArgs, ConfigError};
+pub use connect::{connect, https_connector, ConnectionError};
+pub use utils::ReadEnvError;
+
+/// Re-exports of this crate's own dependencies, at the versions it was built with.
+///
+/// `armonik` re-exports these in turn rather than declaring its own version requirements for the same
+/// crates, so nothing in the workspace can resolve to two incompatible copies of `rustls` (or of
+/// `tonic`, or of `hyper`) in one build.
+pub mod reexports {
```
lemaitre-aneo, line 25: There should most likely be serde when the feature is enabled

packages/rust/armonik/src/client/mod.rs
```diff
@@ -408,7 +295,7 @@ where
         self.channel.poll_ready(cx)
     }
 
-    fn call(&mut self, request: hyper::http::Request<tonic::body::Body>) -> Self::Future {
+    fn call(&mut self, request: tonic::codegen::http::Request<tonic::body::Body>) -> Self::Future {
```
lemaitre-aneo, line 300: I am not sure about this change. What was the reason for it?

packages/rust/armonik/src/client/mod.rs
```diff
@@ -71,134 +69,23 @@ pub struct Client<T = tonic::transport::Channel> {
 impl Client<tonic::transport::Channel> {
     /// Create a new client using the configuration from the environment variables
     pub async fn new() -> Result<Self, ConnectionError> {
-        Self::with_config(ClientConfig::from_env().context(ConfigSnafu {})?).await
+        Self::with_config(ClientConfig::from_env()?).await
```
lemaitre-aneo: You lose the "simplified stacktrace" (based on location) with that. I think Snafu types should be public and hidden to let other crates use them if needed, like here.

packages/rust/armonik/src/lib.rs
```diff
@@ -14,14 +14,13 @@ pub use objects::*;
 mod utils;
 
 pub mod reexports {
```
lemaitre-aneo, line 19: armonik-transport should be reexported as well. Maybe as a root module named transport instead of reexports, though

wkirschenmann:
All five addressed in 7a45c510, as a new commit rather than an amend.

`serde`, `serde_json` and `serial_test` are workspace dependencies now, and `reexports` exposes `serde` under
the feature.

On the snafu location: you were right, and I took the option you suggested. The selectors are `pub` and
`#[doc(hidden)]`, the `From` impl is gone, and `Client::new` builds the error itself again with
`.context(ConfigSnafu {})`. One detail worth noting: snafu keeps generated selectors module-private
regardless of the enum's visibility, so it needed `#[snafu(visibility(pub))]` on `ConnectionError`.

`armonik-transport` is re-exported as a root `transport` module, as you suggested.

On `mod.rs:298`: `hyper` was a direct dependency of `armonik` only for the connection code that moved out.
`tonic::codegen::http` is the same `http` crate, reached through `tonic`, which guarantees it is the version
`tonic` itself uses; keeping `hyper` would have meant a whole dependency for one type alias. It stays a
dev-dependency, for the `get_nb_request` test helper. Happy to restore it if you would rather see the smaller
diff.

packages/rust/armonik/Cargo.toml
```diff
@@ -1,62 +1,51 @@
...
-  "tls12",
-], default-features = false, optional = true }
-tokio = { version = "1.52", default-features = false, optional = true }
+snafu.workspace = true
+tracing.workspace = true
+tracing-futures = { workspace = true, features = ["futures-03"] }
+tokio = { workspace = true, optional = true }
 serde = { version = "1.0", features = [
   "derive",
 ], default-features = false, optional = true }
```
lemaitre-aneo: use workspace

packages/rust/armonik-transport/Cargo.toml
wkirschenmann: Done in 497f4bd6. `serde` is `serde = { workspace = true, optional = true }`, and `serde_json` and `serial_test` moved to the workspace too.

packages/rust/armonik-transport/src/lib.rs
wkirschenmann: Done in 497f4bd6: `#[cfg(feature = "serde")] pub use serde;` in `reexports`.

packages/rust/armonik/src/client/mod.rs
wkirschenmann: `hyper` was a direct dependency of `armonik` only for the connection code that moved to `armonik-transport`. `tonic::codegen::http` is the same `http` crate, reached through `tonic`, so it is guaranteed to be the version `tonic` itself uses; keeping `hyper` would mean a whole dependency for one type alias. It stays a dev-dependency for the `get_nb_request` test helper. Happy to restore it if you prefer the smaller diff.

packages/rust/armonik/src/client/mod.rs
wkirschenmann: Agreed, and done your way in 497f4bd6: the `From` impl is gone, the selectors are `pub` and `#[doc(hidden)]`, and `Client::new` builds the error itself again with `.context(ConfigSnafu {})`. One detail worth knowing: snafu keeps its generated selectors module-private regardless of the enum visibility, so it also needed `#[snafu(visibility(pub))]` on `ConnectionError`.

packages/rust/armonik/src/lib.rs
wkirschenmann: Done in 497f4bd6, as you suggested: `pub use armonik_transport as transport;` at the root, so `armonik::transport::ClientConfig` resolves.

packages/rust/armonik/Cargo.toml
```diff
@@ -1,62 +1,51 @@
...
 _gen-server = ["tonic/server", "tonic/router", "dep:tokio"]
 
 [dependencies]
-tonic = { version = "0.14", features = ["codegen"], default-features = false }
-tonic-prost = "0.14"
+# TLS, mTLS and the rest of the connection story live in `armonik-transport`; this crate keeps the
+# services and the generated types. By path *and* by version: a path dependency cannot be published, so
+# `cargo publish` rewrites this into the version requirement below, which therefore has to name the
+# version actually being released. `nr update-versions` keeps it in step.
+armonik-transport = { path = "../armonik-transport", version = "3.29.2-beta-0", optional = true }
```
lemaitre-aneo: Should be declared in workspace.

wkirschenmann: Done in 28546e03.

One side effect worth knowing: the version now shares a file with `[workspace.package] version`, and
`verify-versions` keyed its findings by file, so the second silently overwrote the first. It takes an explicit
key now, and a desynchronised version does fail `nr verify-versions`.

packages/rust/armonik/Cargo.toml
wkirschenmann: Done in 28546e03, and it fixed a bug rather than only tidying: the workspace entry carries `std`, which this
declaration lacked, so `cargo build -p armonik --features serde` failed on `String: Serialize` here and still
does on main.

Applied to the rest of the manifest too, which removed a `tokio` declared twice, at 1.52 and at 1.48.

lemaitre-aneo (approved):

PR #690, opened 2026-07-30
fix(rust): apply the request timeout and rate limit that were being parsed and dropped

wkirschenmann:
`GrpcClient__Timeout` and `GrpcClient__RateLimit` were parsed into `ClientConfig` and never passed to
anything: `connect` copied fifteen of the seventeen fields onto the endpoint and skipped these two. Setting
either had no effect, with no error and no warning.

Both are now applied, through `Endpoint::timeout` and `Endpoint::rate_limit`.

## The default that had to change with it

`timeout` parsed an empty value to `Some(60s)` while its doc comment said "defaults to no timeout". Nothing
applied it, so the 60s was never observable, and wiring the field through while keeping that default would
have imposed a one-minute deadline on every request of every existing caller. It is now `None`, as documented.

`connect_timeout` is the mirror image and is left alone: its 60s has always been applied, so there the doc
comment is what was wrong, and only the comment changes.

## Testing

`armonik-transport` had no tests. This adds a fixture serving one gRPC method that sleeps, and six tests:
the timeout firing, a slow call finishing without one, a rate limit not breaking a call, and the three
default and parsing cases.

The first was checked against the bug. With the two lines removed from `connect`,
`cargo test -p armonik-transport --test timeout` fails on the elapsed-time assertion after 10.02s, the call
having completed rather than been cut short.

## Also here

A zero-valued rate limit is refused while parsing. `tower`'s rate limiter asserts both halves are above
zero, and this is what starts handing them to it, so `0/1s` would otherwise panic inside `connect`.

Comments that narrated the code's history are gone, following a review comment on this pull request. The
same habit was in the transport crate's own module docs, which also named `armonik` from inside the layer
below it; corrected here because these are files this change already touches, plus the crate's README.

## Impact

`GrpcClient__Timeout` and `GrpcClient__RateLimit` now do what they say, so anyone who set them and got
nothing will start getting what they asked for. No change for anyone who set neither. No API change; the new
dev-dependencies are for the test fixture only.

### DIFF
--- a/packages/rust/armonik-transport/src/config.rs
+++ b/packages/rust/armonik-transport/src/config.rs
@@ -315,10 +315,23 @@ impl ClientConfig {
                 .context(InvalidRateLimitCountSnafu {
                     value: parts[0].to_string(),
                 })?;
-            let duration = parts[1]
+            let duration: Duration = parts[1]
                 .parse::<humantime::Duration>()
-                .context(InvalidDurationSnafu { value: rate_limit })?
+                .context(InvalidDurationSnafu {
+                    value: rate_limit.clone(),
+                })?
                 .into();
+            // `tower`'s rate limiter asserts both are non-zero, so leaving these to it turns a mistyped
+            // option into a panic inside `connect` rather than an error the caller can read.
+            if limit == 0 || duration.is_zero() {
+                return IncompatibleOptionsSnafu {
+                    msg: format!(
+                        "`GrpcClient__RateLimit={rate_limit}` has a zero count or duration. Both have \
+                         to be above zero, as in `100/1s`; leave it empty for no rate limit"
+                    ),
+                }
+                .fail();
+            }
             Some((limit, duration))
         };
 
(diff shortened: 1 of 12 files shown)


packages/rust/armonik-transport/src/connect.rs
```diff
@@ -32,6 +34,13 @@ pub async fn connect(config: ClientConfig) -> Result<tonic::transport::Channel,
         transport_endpoint = transport_endpoint.origin(target);
     }
 
+    if let Some(timeout) = timeout {
+        transport_endpoint = transport_endpoint.timeout(timeout);
+    }
+    if let Some((limit, duration)) = rate_limit {
+        transport_endpoint = transport_endpoint.rate_limit(limit, duration);
```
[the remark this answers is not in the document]
wkirschenmann: Correct, and reproduced before fixing. Done in 2414e273.

Measured by driving `connect`: `0/1s` panicked at `tower-0.5.3/src/limit/rate/rate.rs:17`, `1/0s` at line
18, both after the parsing had accepted them. Both halves are now rejected while parsing, with a message
that names the option and quotes the value.

packages/rust/armonik-transport/src/config.rs
```diff
@@ -287,8 +287,11 @@ impl ClientConfig {
             )
         };
 
+        // `None`, not `Some(60s)`: nothing applied this until now, so the 60s was never observable, and
+        // keeping it while starting to apply it would deadline every existing caller's requests at one
+        // minute. `connect_timeout` above is the opposite case, its 60s having always been applied.
```
lemaitre-aneo: No need to state the history of this code.

wkirschenmann: Removed.

lemaitre-aneo (approved):

PR #692, opened 2026-07-30
ci(rust): refuse a committed copy where a symlink belongs, and document the Windows case

wkirschenmann:
# Motivation

`packages/rust/armonik` reaches the shared `Protos` directory, `README.md` and `LICENSE` through symlinks.
That is deliberate: the repository keeps one copy of each, and `cargo package` follows the links to vendor
them into the published crate.

Creating a symlink on Windows normally requires `SeCreateSymbolicLinkPrivilege`, which an ordinary unelevated
account does not hold. Git checks while cloning and, finding it absent, records `core.symlinks = false` in the
repository's own `.git/config` — after which each entry is checked out as an ordinary file containing its
target, such as the fifteen bytes `../../../Protos`.

Developer Mode gets a contributor out of this, and the second commit here documents it accurately after
measuring what it actually does: it does *not* grant the privilege, it lets `CreateSymbolicLink` accept a flag
waiving the check, which the calling program must pass. Git passes it; PowerShell 5.1's
`New-Item -ItemType SymbolicLink` does not, and still fails on a machine where Git works — which reads as if
Developer Mode had not taken effect. The step that actually matters is `git config --unset core.symlinks`: a
repository-local `false` beats a global `true`, so setting the global alone changes nothing.

The Rust crate then does not build there, since `protos/V1` is not a directory. The obvious local fix is to
delete the placeholder and copy in what it pointed at — which works, and which would be a slow disaster if
committed:

- the crate's protos would drift from the shared `Protos` with nothing to notice the divergence;
- and a crate packaged from such a checkout ships **no protos at all**. `include = ["protos/**"]` matches
  nothing when `protos` is a file rather than a directory. Confirmed rather than assumed:

  ```
  $ cargo package --list --allow-dirty | grep proto
  (nothing)
  ```

  A consumer of that published crate could not build it.

So the local copy needs to be possible, and committing it needs to be impossible.

# Description

**`scripts/check-symlinks.sh`** asserts, for each of the three paths, that the index holds exactly one entry,
of mode `120000`, pointing where it should. Run from the repository root; annotates the offending path when
run in CI, and points at `CONTRIBUTING.md`.

All three, not only `protos`: `README.md` and `LICENSE` have the same failure mode, and nobody had noticed
them. `protos` is the one that breaks the build; the other two break `cargo package` quietly, shipping a
crate whose README and licence are fifteen-byte relative paths.

**A `Check symlinks` job** in `ci.yml` runs it on every pull request. Its own job rather than a step in an
existing one, so that a red cross names the problem.

**`CONTRIBUTING.md` gains a Windows section** covering both routes:

- Enabling symlinks through Developer Mode, which needs no elevation, with the `git config --unset
  core.symlinks` that is easy to miss and the note about which tools pass the flag.
- Working with a local copy, with the reason not to commit it and a pointer to the check that refuses it.

# Testing

The check was verified against both ways it can actually go wrong, by constructing the failing index in a
scratch `GIT_INDEX_FILE` rather than by reading the code:

| index state | result |
|---|---|
| the repository as it is | `exit 0`, three lines listing the links and their targets |
| a symlink replaced by an ordinary file — what a Windows contributor would commit | `exit 1`, error annotated on the path |
| a symlink replaced by a copy of the tree it pointed at | `exit 1` |

The first version of the script **passed** that third case: with the path gone from the index, the loop had
nothing to iterate over and reported nothing. That is why it now asserts presence before mode. Worth
mentioning because it is the failure this check exists for, and it was the one it missed.

The script is committed mode `100755`, matching the other scripts in `scripts/`, since the workflow invokes
it as `./scripts/check-symlinks.sh`. That also needed catching: a Windows checkout does not carry the
executable bit, so it committed as `100644` and CI would have failed with "Permission denied".

`ci.yml` parses, and the job resolves as expected.

# Impact

- One new CI job, a few seconds, no dependency on any toolchain.
- No behaviour change to any package. Nothing but a workflow, a script and a documentation section.
- A Windows contributor now has a documented way to build, and a guard rail if they take it.
- Existing contributors see nothing new unless they replace one of the three paths, which is the point.

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas. *(the script's header explains why
      the symlinks exist and why a copy must not be committed, since that is the part a future reader will
      question)*
- [x] I have made corresponding changes to the documentation. *(the new Windows section in
      `CONTRIBUTING.md`, which is what the check's error message points at)*
- [x] I have thoroughly tested my modifications and added tests when necessary. *(see Testing — both failure
      modes exercised against a real index)*
- [x] I have assessed the performance impact of my modifications. *(a few `git ls-files` calls in a job of
      its own)*

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,25 @@ jobs:
       - name: Verify Versions
         run: nr verify-versions
 
+  check-symlinks:
+    name: Check symlinks
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          fetch-depth: 0
+
+      - name: The repository's symlinks must still be symlinks
+        run: |
+          # The script reports to stderr and answers with its exit status; the annotation is added here
+          # rather than by the script, so that the script stays useful to a contributor running it by hand
+          # as CONTRIBUTING.md suggests.
+          if ! ./scripts/check-symlinks.sh; then
+            echo "::error::a path that must be a symlink is not one; see the job log and the Windows section of CONTRIBUTING.md"
+            exit 1
+          fi
+
   format-csharp:
     name: Format C#
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,14 +41,7 @@ jobs:
           fetch-depth: 0
 
       - name: The repository's symlinks must still be symlinks
-        run: |
-          # The script reports to stderr and answers with its exit status; the annotation is added here
-          # rather than by the script, so that the script stays useful to a contributor running it by hand
-          # as CONTRIBUTING.md suggests.
-          if ! ./scripts/check-symlinks.sh; then
-            echo "::error::a path that must be a symlink is not one; see the job log and the Windows section of CONTRIBUTING.md"
-            exit 1
-          fi
+        run: ./scripts/check-symlinks.sh
 
   format-csharp:
     name: Format C#
--- a/scripts/check-symlinks.sh
+++ b/scripts/check-symlinks.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+#
+# Verify that the symlinks this repository relies on are still symlinks.
+#
+# `packages/rust/armonik` reaches the shared `Protos`, `README.md` and `LICENSE` through symlinks, which is
+# what lets `cargo package` vendor them into the published crate while keeping a single source in the
+# repository.
+#
+# Git only materialises a symlink when `core.symlinks` is enabled, and that is not the default on Windows:
+# there each one is checked out as an ordinary file containing its target. Committing such a file — or a
+# copy of the tree it pointed at — would detach the crate from the shared source and let the two drift, so
+# this refuses it.
+#
+# Run from the repository root, in CI or by hand. Diagnostics go to stderr; the exit status is the verdict.
+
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -31,6 +31,25 @@ jobs:
...
+
+      - name: The repository's symlinks must still be symlinks
+        run: |
+          # The script reports to stderr and answers with its exit status; the annotation is added here
+          # rather than by the script, so that the script stays useful to a contributor running it by hand
+          # as CONTRIBUTING.md suggests.
+          if ! ./scripts/check-symlinks.sh; then
+            echo "::error::a path that must be a symlink is not one; see the job log and the Windows section of CONTRIBUTING.md"
+            exit 1
+          fi
```
lemaitre-aneo: Just call the script. The description of the script is enough, and the script already prints out what the issue is.

scripts/check-symlinks.sh
lemaitre-aneo, line 1: I feel that the script could be made POSIX compliant fairly easily

wkirschenmann:
Both done in a750eda5.

The step is now just `run: ./scripts/check-symlinks.sh`.

The script is POSIX: `#!/bin/sh`, `[` instead of `[[`, `set -u` without `pipefail`, no `local`. Verified with
`sh -n` and `dash -n`, and both failure modes re-exercised against a scratch index.

One thing to flag: `[[` was SonarCloud's own suggestion on this PR (`shelldre:S7688`), and it is bash-only, so
the two cannot both be satisfied. I went with POSIX. If Sonar raises it again, that is why.

lemaitre-aneo (approved):

PR #694, opened 2026-07-31
build(rust): declare the build script's proto dependencies

wkirschenmann:
`packages/rust/armonik/build.rs` declares no `cargo:rerun-if-changed`. With none declared, cargo re-runs the
build script whenever anything in the crate changes, so editing a test file re-runs `protoc` over all forty
protos and rebuilds the library.

Measured from a settled baseline:

|                        | before | after |
|------------------------|--------|-------|
| touch a `.proto`       | rebuilds | rebuilds |
| touch `tests/agent.rs` | rebuilds the library | does not |

The forty file names become a `const` array joined onto a `PROTO_ROOT`, because the paths are needed twice
now: once for the `rerun-if-changed` lines, once for `compile_protos`.

Nothing else changes. Same protos, same options, identical generated code.

Note that declared dependencies replace cargo's fallback rather than adding to it, so `PROTO_FILES` is now
load-bearing for rebuild decisions as well as for compilation. A proto missing from that list was never
compiled anyway.

Verified with `cargo build --all --locked`, `cargo fmt --all --check`, `cargo doc` and
`cargo clippy --all --no-deps -- -Dwarnings -Dunused-crate-dependencies`, all clean.

### DIFF
--- a/packages/rust/armonik/build.rs
+++ b/packages/rust/armonik/build.rs
@@ -1,7 +1,7 @@
 /// The proto files to compile, relative to [`PROTO_ROOT`].
 ///
-/// A list rather than a glob, so that adding a proto stays a deliberate act — as it already was. The only
-/// change is that the `protos/V1/` prefix is no longer repeated forty times.
+/// A list rather than a glob, as it already was; the only change is that the `protos/V1/` prefix is no
+/// longer repeated forty times.
 const PROTO_FILES: &[&str] = &[
     "agent_common.proto",
     "agent_service.proto",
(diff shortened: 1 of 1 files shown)


lemaitre-aneo:
This is not a perf PR

wkirschenmann:
Agreed, it is not. Retitled to `build(rust): declare the build script's proto dependencies`, which is what it
actually is: a change to the build system, not to anything at run time.

The commit message inside still says `perf(rust)`. I left it, since rewriting a pushed commit removes the
review history attached to it; the PR title is what the semantic check reads and what a squash merge would
use.

I also cut the description down, which was far too long.

lemaitre-aneo (approved):

PR #697, opened 2026-08-01
feat(rust): honour ReusePorts, and let the Rust tests run on Windows

wkirschenmann:
> Depends on #696, which is its base branch.

`GrpcClient__ReusePorts` did nothing here. It is what keeps a Windows client from exhausting the
ephemeral port range when it opens many connections in a short window, and it is why the Rust legs of
`test.yml` carried `# FIXME: Rust should be working on Windows, but we most likely encounter a port
exhaustion issue`. That exclusion is removed here.

## Description

`SO_REUSE_UNICASTPORT` has to be set before `connect`, so socket creation cannot be left to hyper.
`TcpConnector` keeps hyper's connector for every other case, and only a Windows caller who asks for the
option reaches the one written here. The `setsockopt` call goes through `windows-sys`, so its signature
and both constants come from the SDK metadata rather than being copied by hand.

Owning socket creation means owning what hyper does around it, and one thing was missing: it races
addresses with a 300ms fallback, so a dual-stack host with one blackholed record still connects quickly.
`race` does the same. Without it, turning the option on would have traded a 300ms connection for a
`connect_timeout`-long one.

The default is `true`, matching the C# client, whose documentation defines this option as "Enable the
option SO_REUSE_UNICASTPORT upon socket opening".

## Testing

`cargo test -p armonik-transport --all-features` gives 89 tests, all passing. Six drive the racing policy
on a paused clock, three Windows-only ones pin what Winsock accepts, and three open a hundred channels at
once with the option on and off.

The scheduling is tested apart from the sockets deliberately: there is no way to make a socket blackhole
a `SYN` from a test, which is why hyper's own equivalent test needs real unreachable addresses, is
`#[ignore]`d twice and does not run in their CI.

## Impact

**The default changes for Windows callers**, who now get the option and the connector that sets it. That
is the alignment with the other clients, and `GrpcClient__ReusePorts=false` opts out. Nothing changes on
any other platform, down to which connector is used.

The Rust legs now also run on `windows-latest`, with nothing added about the symlinks a Windows checkout
may not create: both legs pass as they are, so those runners need no help there.

Three new direct dependencies: `windows-sys` under `cfg(windows)` for the socket option, `socket2` for
keepalive on a socket this crate opens itself, and `futures-util` for holding several connection attempts
in flight.


packages/rust/armonik-transport/src/tcp.rs
```diff
@@ -0,0 +1,607 @@
...
+/// Generic over how an address is reached so that the scheduling can be tested on its own. The real
+/// thing is `connect_to`, and there is no way to make a socket blackhole a `SYN` from a test.
+async fn race<T, A, F>(
+    addresses: Vec<std::net::SocketAddr>,
+    fallback: Duration,
+    attempt: A,
+) -> Result<T, BoxError>
+where
+    A: Fn(std::net::SocketAddr) -> F,
+    F: Future<Output = Result<T, BoxError>>,
```
lemaitre-aneo, line 47: This could most likely be made generic and implemented in utils, but I wonder if it would be worth it.

packages/rust/armonik-transport/src/tcp.rs
```diff
@@ -0,0 +1,607 @@
...
+/// Await `body`, bounded by an absolute `deadline` when one is set. [`None`] means it expired.
+async fn with_optional_deadline<T>(
+    deadline: Option<tokio::time::Instant>,
+    body: impl std::future::Future<Output = T>,
+) -> Option<T> {
+    match deadline {
+        Some(deadline) => tokio::time::timeout_at(deadline, body).await.ok(),
+        None => Some(body.await),
+    }
+}
```
lemaitre-aneo: Why would you return an Option instead of the tokio Result directly?

packages/rust/armonik-transport/src/tcp.rs
```diff
@@ -0,0 +1,607 @@
...
+
+/// How the underlying TCP connection is opened.
+#[derive(Debug, Clone)]
+pub enum TcpConnector {
+    /// hyper's connector. The default, and unchanged.
+    Standard(HttpConnector),
+    /// Our own connector, which can defer ephemeral port allocation.
+    ReusePorts(ReusePortsConnector),
+}
```
lemaitre-aneo, line 140: As you need to reimplement everything in ReusePortsConnector, what's the use for keeping HttpConnector around?

packages/rust/armonik-transport/src/config.rs
```diff
@@ -230,6 +238,9 @@ pub struct ClientConfigArgs {
     /// [`Secret`].
     #[cfg_attr(feature = "serde", serde(default))]
     pub proxy_password: Secret,
+    /// Let the OS reuse local ports for outgoing connections, defaults to true
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub reuse_ports: ReusePorts,
```
lemaitre-aneo: It would be better to use Option<bool> and handle default when converting into a ClientConfig

wkirschenmann: Done in 905be71f.

packages/rust/armonik-transport/src/tcp.rs
wkirschenmann: No good reason. It returns the `Result` now, in 905be71f.

packages/rust/armonik-transport/src/tcp.rs
wkirschenmann: Left here for now, with its one caller. Easy to move if a second use emerges.

packages/rust/armonik-transport/src/tcp.rs
wkirschenmann: It reimplements a subset, not everything: no local address or interface binding, no pluggable resolver. And off Windows the option does nothing, so routing every caller through ours would move them onto newer code for no gain. #703 deletes ours once `HttpConnector` gains the setter upstream.

packages/rust/armonik-transport/src/tcp.rs
```diff
@@ -0,0 +1,607 @@
...
+            .map_err(|source| BoxError::from(format!("could not resolve `{host}`: {source}")))?
+            .collect::<Vec<_>>();
+
+        if addresses.is_empty() {
+            return Err(BoxError::from(format!("`{host}` resolved to no address")));
+        }
+
+        let attempts = race(addresses, FALLBACK_DELAY, |address| {
+            self.connect_to(address)
+        });
```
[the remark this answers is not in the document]
wkirschenmann: Correct, and the 600ms is exactly what the raw order gave. Interleaved by family now, in c1ed377f.

packages/rust/armonik-transport/src/tcp.rs
lemaitre-aneo: Then, I am not comfortable with having a sub-par variant that is the default on Windows.

wkirschenmann:
Set to draft to prevent merge.
We will need to have the setter in hyper-utils be pushed and integrated before we can have a good inplementation here. Still this PR will serve as a base for other developements.

packages/rust/armonik-transport/src/tcp.rs
wkirschenmann: Would you consider a feature or config guard enough ? With a warning log to prevent production usage ?
It wouls at least allow the windows CI to run.

wkirschenmann:
Closed pending re-authoring on the new base. This branch sits on the pre-recut lineage, and its content is re-authored as a unit declaration on top of the config machinery once that lands.

The branch is retained as the source material - nothing here is lost, and a new PR opens against the re-cut stack.

PR #704, opened 2026-08-02
refactor(rust): hand the transport the certificates rather than paths to them

wkirschenmann:
> Depends on #701, which is its base branch.

# Motivation

`from_config_args` called `std::fs::read_to_string` on the paths in `cert_pem`, `key_pem` and
`ca_cert`. Opening a file is integration with a deployment, the same as reading an environment
variable, which #701 already moved out of the transport. Knowing where material is kept belongs to the
layer that knows what a deployment looks like.

It is also what the C ABI needs: a caller converting a PKCS#12 file, or reading the Windows
certificate store, has bytes and no path, and should not have to write a private key to disk to be
understood.

# Description

The three options carry the PEM itself. Their names already said `pem`, so only their meaning moves.
The transport calls no `std::fs`, and `ConfigError::Io` goes with it. `armonik` opens the files that
`GrpcClient__CertPem`, `GrpcClient__KeyPem` and `GrpcClient__CaCert` name, through `read_pem_file`;
`EnvConfigError` gains a `File` variant naming both the variable and the path.

`key_pem` becomes a `Secret`, as `proxy_password` is: a private key now sits in a struct that derives
`Debug` and `Serialize`. The compiler found the first consequence by itself, which is the point of the
type. The tracing span recorded `cert_pem` and `key_pem`, harmless while they were paths and not once
they are the material. It records whether each was supplied, and nothing else.

# Testing

`cargo test -p armonik-transport --all-features` gives 89 passing, and
`cargo test -p armonik --all-features --lib client::env` 8.

The two tests that asserted a missing file follow the behaviour to `armonik`, where they exercise
`read_pem_file`. It takes the variable name, so they borrow one of their own rather than a
`GrpcClient__*` that the mock-backed tests in the same binary depend on. In their place the transport
pins what a caller still passing a path now gets, for all three options: a PEM error, rather than this
crate quietly opening whatever it points at.

# Impact

Nothing changes for a deployment. The variables still name files, and `armonik` still reads them.

For a Rust caller building `ClientConfigArgs` by hand, the three fields change meaning, and `key_pem`
changes type. Passing a path is caught as an invalid certificate rather than silently read.

# Additional Information

This is what lets the C ABI take its whole configuration as one JSON document, certificates included,
instead of needing a second channel for them.

# Checklist


lemaitre-aneo (commented):
You can always access ClientConfig directly and set the certificates there. In the path you mentioned, the certificates are read during the conversion from ClientConfigArgs to ClientConfig.
If you really want to expose certificate content in ClientConfigArgs, best way would be to have both cert_pem_path and cert_pem_content, and chose which to use during the conversion.

wkirschenmann:
Superseded by #705: same Certificate<T> design, rebuilt on top of #696 directly (with the generic FromEnv mechanism it needed), rather than the #701/#697 chain this branch went through.

PR #707, opened 2026-08-03
feat(rust): give the transport the retry policy, and the caller the loop

wkirschenmann:
# Motivation

1. Replay a request the gRPC retry convention already covers, so a transient failure does not
   surface to the caller.
2. Match the default policy the other ArmoniK clients use, so a deployment behaves the same
   whichever client is talking to it.

# Description

`RetryPolicy` carries five attempts, one second growing by 1.5 to a five-second ceiling, replaying
on `Unavailable`, `Aborted` and `Unknown` - what the other clients already do by default. The wait
follows the gRPC spec, `random(0, min(initial * multiplier^n, max))`. Two independent bounds cap how
much of a request may still be held for a replay: `max_buffer_per_call` for a stream, `max_unary_size`
for a single message.

The `retry!` macro expands in the caller's own function so each attempt can borrow the client it is
made on. `GrpcStatus` reads a gRPC code out of whatever error type an attempt produces, implemented
for `tonic::Status`. No caller in this repository invokes the macro yet, so the options change
nothing until one does; that is deliberate, the loop belongs to whoever knows a method is unary and a
request can still be reproduced.

`ClientConfig` gains a `retry: RetryPolicy` field; `ClientConfigArgs` gains the matching seven string
options (`max_attempts`, `initial_backoff`, `max_backoff`, `backoff_multiplier`,
`retryable_status_codes`, `max_retry_buffer_per_call`, `max_retry_unary_size`), already covered by
`#[derive(FromEnv)]` - no line written for them. A parse failure on any of the seven names its own
field rather than a fixed one, which `InvalidRetryOption` had gotten wrong for all of them the same
way #705 (this branch's base) had for `InvalidDuration`/`InvalidInteger`/the proxy messages.

# Testing

`cargo test --workspace --all-features --no-fail-fast` (minus `client::`, run separately): 169
passed. `cargo fmt --all --check`, CI's exact clippy invocation, and `cargo doc` with warnings as
errors: clean. See the last commit for the exact commands.

Coverage: the backoff sequence and its ceiling, a retryable failure retried until attempts are
spent, a non-retryable or status-less failure ending at once, `max_attempts` of 1 never replaying,
abandoning the call abandoning the wait, and each option read from the environment or naming itself
when it fails to parse.

# Impact

Nothing changes for a deployment that leaves the options alone: the defaults match what the other
clients already do, and the macro is not called anywhere yet. A caller that wants retries writes
`armonik_transport::retry! { ... }` around its own call.

# Additional Information

Stacked on #705 (`wk/refactor/rust-certificates-v2`), which this branch is rebased onto: the
`Certificate`/`FromEnv` work it depends on is not merged yet.

# Checklist


wkirschenmann:
Closed pending re-authoring on the new base. This branch sits on the pre-recut lineage, and its content is re-authored as a unit declaration on top of the config machinery once that lands.

The branch is retained as the source material - nothing here is lost, and a new PR opens against the re-cut stack.

PR #708, opened 2026-08-06
refactor(rust): read the environment through figment, and rename ClientConfig to HttpConfig

wkirschenmann:
First of three PRs replacing #705, split after its base (#696) moved and left it conflicting.
This one: adopt figment for environment reading, then rename ClientConfig(Args) to
HttpConfig(Args) ahead of a GrpcConfig a later PR introduces for tonic/retry settings.

- Replaces the hand-rolled deserializer (which could not support `#[serde(flatten)]`, needed
  by the next PR's thematic-unit grouping) with figment's `Env` provider, plus a `text`/
  `secret_text` deserialize_with shim so every field still carries plain text regardless of
  figment's scalar coercion.
- Retires `HttpConfig::from_env` (the resolved type holds a `Uri`/`CertificateDer`/
  `PrivateKeyDer`, none of which has a `Deserialize` worth offering). `armonik::client::env`
  now composes `HttpConfigArgs::from_env` + `HttpConfig::from_config_args` in `Client::new`,
  which returns a new `NewClientError` instead of `ConnectionError` - a breaking change, but
  this crate has not been published with either shape yet.

Stacked as: this PR -> wk/refactor/rust-config-units -> wk/feat/rust-pkcs12.


wkirschenmann:

PR #709, opened 2026-08-06
refactor(rust): group the config's flat fields into thematic units

wkirschenmann:
Second of three PRs replacing #705 (see wk/refactor/rust-config-figment for the first). Groups
the flat config fields into thematic units, each with its own `{Unit}ConfigArgs`/`{Unit}Config`
pair and `resolve()`, in a new file per unit.

- `tcp_config.rs`, `http2_config.rs`: grouped via `#[serde(flatten)]` +
  `serde_with::with_prefix!`, since their fields already share one literal prefix (`Tcp*`,
  `Http2*`). Every environment variable and JSON key stays exactly as it is today.
  `parse_optional_duration`/`parse_optional_int` land in `config.rs` as shared helpers,
  introduced here and reused by every field that needs "empty means None, else parse" from
  now on, so nothing later has to deduplicate copy-pasted parsing.
- `proxy_config.rs`: `ProxyConfigArgs { source, username, password }` in its final shape from
  the start, each field individually renamed since they do not share one prefix. Adds the
  `EnvGuard` RAII helper to `tests/common/mod.rs` so a proxy env-var test that panics between
  `set_var` and `remove_var` cannot leak state into the next `serial(env)` test.
- `tls_config.rs`: `TlsConfigArgs`/`TlsConfig`, grouped via plain `#[serde(flatten)]`. Fixes a
  real bug while moving this code: `override_target_name` that failed to parse named the
  endpoint in its error instead of the value that actually failed. `ConfigError::Uri` and
  `::Http` merge into one `Uri` variant parameterized by `option: &'static str`, the same
  convention every other variant already follows. `TlsConfig` gets a hand-written `Clone`
  (`PrivateKeyDer` has none, only `clone_key()`); `HttpConfig` switches from its own
  hand-written `Clone` to `#[derive(Clone)]` in the same commit, now that every field is
  `Clone` on its own. No PKCS12 yet - that is the third PR.

Stacked as: wk/refactor/rust-config-figment -> this PR -> wk/feat/rust-pkcs12.

### DIFF
--- a/packages/rust/Cargo.lock
+++ b/packages/rust/Cargo.lock
@@ -12,6 +12,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ae221649c9976a6f6c56ae1facf410f3ddb33cc661c4b7b61020a912d4237fbc"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -60,6 +69,7 @@ dependencies = [
  "secrecy",
  "serde",
  "serde_json",
+ "serde_with",
  "serial_test",
  "snafu",
  "tokio",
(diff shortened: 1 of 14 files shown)


wkirschenmann:

PR #710, opened 2026-08-06
feat(rust): load a client certificate from a PKCS12 bundle

wkirschenmann:
Third of three PRs replacing #705 (see wk/refactor/rust-config-figment and
wk/refactor/rust-config-units for the first two). Adds PKCS12 as a second way to load the
client's certificate and key.

- `cert_p12` (a bundle path) and `cert_p12_password` (`Secret`) on `TlsConfigArgs`/
  `TlsConfig`, mutually exclusive with `cert_pem`/`key_pem`: the client's certificate and key
  bundled together, the form Windows and most certificate authorities hand out. Read via
  `p12-keystore` and re-encoded as the same DER shapes a PEM pair produces.
  `ConfigError` gains `Pkcs12` and `EmptyPkcs12`, naming the file on a read or parse failure.
- Documents the TLS/mTLS options in the crate README, which never had this section:
  `cert_pem`/`key_pem`/`ca_cert`, `cert_p12`/`cert_p12_password`, `allow_unsafe_connection`,
  `override_target_name`.

### DIFF
--- a/packages/rust/Cargo.lock
+++ b/packages/rust/Cargo.lock
@@ -3,6 +3,41 @@
 version = 4
 
 [[package]]
+name = "aead"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1973cfbc1a2daf9cf550e74e1f088c28e7f7d8c1e1418fb6c9dc5184b7e84c99"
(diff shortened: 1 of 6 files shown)


wkirschenmann:

Issue #703, opened 2026-08-02
Upstream a SO_REUSE_UNICASTPORT setter to hyper-util, and delete our TCP connector


wkirschenmann:
`ReusePortsConnector` exists to set one socket option. `SO_REUSE_UNICASTPORT` has to be applied before
`connect`, and `hyper_util`'s `HttpConnector` owns socket creation with no hook to reach it, so honouring
`GrpcClient__ReusePorts` meant writing a connector.

Everything else it does is hyper's behaviour written again: name resolution, racing the resolved
addresses with a fallback delay, interleaving the address families, carrying the keepalive settings over.
That is a maintenance surface we own for one `setsockopt`.

`HttpConnector` already has `set_reuse_address`. A `set_reuse_unicast_port` beside it, under
`#[cfg(windows)]`, would let us delete our connector entirely and go back to the standard one.

Done when the setter is released upstream and `TcpConnector` has a single variant again.

wkirschenmann:
## Why the Rust client carries no `ReusePorts` option, and why no tripwire pins it

Recorded while re-cutting the Rust config work, so this does not have to be rediscovered.

**Where it stands.** The re-cut stack implements no `ReusePorts` option and contains no TCP
connector of its own. The earlier attempt (#697, closed) added both; none of it survives in the
current lineage, so there is nothing in the tree to delete when this is fixed.

**Why it cannot be implemented against `hyper-util` as shipped.** `HttpConnector` exposes no hook to
touch the socket between creation and `connect`. It has setters for the specific options it chose to
support (`set_nodelay`, `set_keepalive`, `set_local_address`, `set_connect_timeout`), but no general
"configure the socket" callback and no `SO_REUSE_UNICASTPORT` setter. So implementing the option
means one of:

- a setter upstream, which is what this issue asks for, or
- a TCP connector written in this crate and used underneath the proxy connector, which is what #697
  did and what the upstream fix would let us delete.

**Why no tripwire test pins it, unlike the `Tunnel` defects.** The tripwires for hyper-util's tunnel
bugs work because each asserts an *observable behaviour* that a fix changes, so the fix turns the
test red. This gap has no such behaviour: it is a missing API, not a wrong result. The only
detection available is a compile-time probe on the method a future release might add, and its name
is unknown - `set_reuse_unicast_port`, `set_reuse_address`, a socket-configuring closure, or
something else. A probe on a guessed name would sit there never firing and read, to anyone who found
[...]

Issue #702, opened 2026-08-02
Contribute our CONNECT tunnel fixes to hyper-util


wkirschenmann:
The Rust transport carries its own HTTP `CONNECT` tunnel rather than using
`hyper_util::client::legacy::connect::proxy::Tunnel`, for two defects in hyper-util 0.1.20, the latest
published version. Both are ours to fix upstream rather than to work around forever.

**A status line split across reads fails a good tunnel.** On the first read, `tunnel.rs:216`:

```rust
let recvd = &buf[..pos];
if recvd.starts_with(b"HTTP/1.1 200") || recvd.starts_with(b"HTTP/1.0 200") { ... }
else if recvd.starts_with(b"HTTP/1.1 407") { return Err(TunnelError::ProxyAuthRequired); }
else { return Err(TunnelError::TunnelUnsuccessful); }
```

`"HTTP/1.1 2"` matches neither prefix and lands in the last arm, so a successful `CONNECT` is rejected
because the bytes arrived in two segments. Legal, and likelier with a slow proxy or a small MSS.
hyperium/hyper-util#300 fixes this. It is open, unmerged and unreleased; follow it rather than duplicate
it.

**Only `200` opens the tunnel.** RFC 9110 says any 2xx switches a `CONNECT` connection to tunnel mode. A
proxy answering `201` gets `TunnelError::TunnelUnsuccessful`. Nothing upstream covers this; we should
open the pull request.

Ours reads to the blank line before parsing and accepts any 2xx, and both behaviours are covered by
integration tests against a real proxy.

Done when both are released upstream. Deleting ours is then tracked by #699.

## Measured, not just read

Driving `Tunnel` against a fake proxy, with an unsplit `200` as the control:

```
plain 200: ACCEPTED, as it should be
201: REFUSED with `tunnel error: unsuccessful`
split: REFUSED with `tunnel error: unsuccessful`
```

The split case writes `HTTP/1.1 2`, waits 50ms, then the rest. Both defects are on hyper-util's main
branch today, not only in the published 0.1.20.

PR #715, opened 2026-08-06
test(rust): pin hyper-util's tunnel defects as tripwires

wkirschenmann:
# Motivation

`hyper_util`'s `Tunnel`, which the proxy connector drives, gets four cases wrong in 0.1.20. A
release that fixes one should turn CI red rather than let the fix pass unnoticed, so the README's
"Known issues" section stays honest.

# Description

Three tripwire tests in `tests/upstream_tunnel.rs` assert that a known upstream defect is still
there: only an exact 200 opens the tunnel where RFC 9110 says any 2xx, an `HTTP/1.0 407` is not
recognised as a request for credentials, and a portless target is dialled on 443 whatever its
scheme. Each failure message says what to do when it fires. The fourth defect, a status line split
across reads being rejected, is documented in the README only: pinning it needs timing assumptions
that make a test flaky on a loaded runner.

The `CONNECT` proxy fixture moves to `tests/common`, shared between the main suite and the
tripwires rather than written twice.

# Testing

Adds 4 tests: the three tripwires and the control proving the fixture lets an ordinary 200
through. `cargo test -p armonik-transport --all-features` passes at this branch's head; the run
is in the commit message.

# Impact

A `hyper-util` version bump can turn CI red by design; the failure message names the tripwire and
the README section to update. No production code is touched.

# Additional Information

Stacks on `wk/feat/rust-proxy-config`; merge that first. The series works without this PR: reject
it alone if pinning upstream defects in the suite is unwanted.

# Checklist

- [ ] Tests pass locally and in the CI. (locally yes; CI runs on this PR)

### DIFF
--- a/packages/rust/armonik-transport/README.md
+++ b/packages/rust/armonik-transport/README.md
@@ -29,6 +29,26 @@ capitalised: this crate accepts any casing, C# treats `NONE` as a proxy URL. And
 URL, which C# passes to the runtime, is refused here: the `CONNECT` handshake would go out in the
 clear to a proxy expecting TLS.
 
+### Known issues
+
+The tunnel handshake is `hyper_util::client::legacy::connect::proxy::Tunnel`, not code written in
+this crate, and as shipped in `hyper-util` 0.1.20 it gets four cases wrong. None are ours to fix
+directly; [hyperium/hyper-util#300](https://github.com/hyperium/hyper-util/pull/300) tracks the
+first two upstream. All but the second are pinned by a tripwire in `tests/upstream_tunnel.rs`, so a
+`hyper-util` release that fixes one turns that test red rather than letting the fix pass unnoticed.
+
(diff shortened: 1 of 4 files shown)


wkirschenmann:
Dropped. The tripwire policy is not worth a PR of its own here.

What it held: four tests pinning `hyper-util` `Tunnel` defects, and a README "Known issues" section describing them. None of that information is lost - the defects are tracked in #702, #720 and #721, which is where they belong. What goes is the in-crate documentation and the tests that would have turned red when a fix landed.

The reuse-ports half that was going to be folded in here is documented on #703 instead: there is no observable behaviour to assert on, only a missing API whose future name is unknown, so a tripwire would have sat there never firing.

The branch is deleted; the commits stay reachable at `refs/pull/715/head` if the tripwires are ever wanted back.

PR #723, opened 2026-08-06
refactor(rust): rename ClientConfig to HttpConfig

wkirschenmann:
# Motivation

`ClientConfig` configures the HTTP transport of a connection, not a client: the `armonik` crate's
`Client` is built on top of it, and the name kept suggesting the two were the same layer. The
rename is also isolated here on purpose, so the configuration re-work that follows reviews without
rename noise.

# Description

Purely mechanical: `ClientConfig` becomes `HttpConfig` and `ClientConfigArgs` becomes
`HttpConfigArgs`, across both crates, doc comments and README included. `rustls::ClientConfig`, a
different type, stays as it is. The diff is symmetric, +68/-68; no identifier other than the two
type names changes.

# Testing

Adds no tests: nothing behaves differently. `cargo test -p armonik-transport --all-features`
passes at this branch's head; the run is in the commit message, with clean clippy, fmt on both
crates, and `cargo check -p armonik`.

# Impact

Breaking rename for both type names; every other identifier, message, and option is untouched.

# Additional Information

Stacks on `wk/test/rust-upstream-tripwires`; merge that first. First of four pieces re-cutting the
configuration re-work.

# Checklist

- [ ] Tests pass locally and in the CI. (locally yes; CI runs on this PR)

### DIFF
--- a/packages/rust/armonik-transport/Cargo.toml
+++ b/packages/rust/armonik-transport/Cargo.toml
@@ -62,7 +62,7 @@ tower-service.workspace = true
 # The boolean options are read straight from the environment, so testing which spellings they accept
 # means setting variables: process-global state, hence one test at a time.
 serial_test.workspace = true
-# Only to round-trip `ClientConfigArgs` in the test that pins the `serde` feature.
+# Only to round-trip `HttpConfigArgs` in the test that pins the `serde` feature.
 serde_json.workspace = true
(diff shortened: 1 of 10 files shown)


wkirschenmann:
Folded into the config machinery PR. The `ClientConfig` -> `HttpConfig` rename lands together with the Args removal rather than as a step of its own, so it does not stand alone in the new series.

The branch is kept for now; it is the tip that carried the rebuilt stack.

PR #728, opened 2026-08-07
feat(rust): describe the option vocabulary as a JSON schema

wkirschenmann:
The flat option vocabulary is becoming an FFI contract: a .NET consumer will
generate its own options type from it. A hand-written mirror would drift from
the code the day someone forgot it, so this derives the description from the
types that already define the vocabulary.

A `schema` feature (schemars 1.x, off by default) makes `HttpConfig` and the
units it embeds describe themselves. One schema, flat, exactly the names
deserialisation reads.

## What is in it

- `schema` feature on `armonik-transport`, pulling in `schemars` and
  `serde_json`. It implies `serde`, and nothing else changes when it is off.
- `examples/generate_schema.rs` prints the schema, for whoever wants a file:
  `cargo run -p armonik-transport --features schema --example generate_schema`.
- The schema function folds into `embed_prefixed!`, so an embedding declares
  its prefix, its reader and its schema in one place. `serde_with::with_prefix!`
  has no schemars integration, hence a helper that reapplies the prefix to the
  generated properties, `required` keys included, inside union branches too.
- `strip_defaults`, because schemars serialises a field's `Default` in the
  field's own type: `false`, or a `Duration` as an object, on an option whose
  schema type is string. Each option states its default in prose instead.
- No committed JSON. Six tests pin the invariants a consumer builds against;
  a schemars upgrade that moves a keyword around breaks nothing that was
  actually promised.

## Two deliberate choices

No mirror structs. The real types derive `JsonSchema`; `TlsConfig` describes
itself through `RawTls`, the shape it already deserialises from, since a unit
built through `TryFrom` describes the shape a document writes rather than the
shape the program keeps.

The schema promises only what the reader reads. The proxy is
`#[serde(skip)]` and no option feeds it, so no proxy option appears; a test
pins that, because an option the schema declares and the reader ignores is a
generated field that silently does nothing.

A unit read under a non-empty prefix also stops spelling an option name in its
field docs. Those docs are now the schema's descriptions, and `Keepalive` is
not a name any source writes - `TcpKeepalive` is, and the prefix belongs to the
embedding. Units under no prefix keep their names, where the two coincide.

## Tests

6 new tests: 5 schema smoke tests, and one in `config_utils` covering the
prefix helper's union-branch case, which no embedding exercises with a
non-empty prefix yet.

## Gates

```
cargo test -p armonik-transport --all-features            80 passed, 0 failed
cargo clippy -p armonik-transport --all-features --all-targets   0 warnings
cargo clippy -p armonik --all-features --all-targets             0 warnings
cargo check -p armonik-transport --no-default-features           0 warnings
cargo fmt --check                                                clean
```

### DIFF
--- a/packages/rust/armonik-transport/src/tls_config.rs
+++ b/packages/rust/armonik-transport/src/tls_config.rs
@@ -2,7 +2,7 @@
 //! verification behaves rather than what is verified.
 //!
 //! Unlike the TCP and HTTP/2 units, these fields share no common prefix in the environment
-//! (`CertPem`, `CaCert`, `AllowUnsafeConnection`, `OverrideTargetName`, ...), so the embedding
+//! (`CertPem`, `CaCertPath`, `AllowUnsafeConnection`, `OverrideTargetName`, ...), so the embedding
 //! composes them with a plain [`serde(flatten)`](serde::Deserialize) and no prefix at all.
 //!
 //! Every option naming a file loads it while the configuration is read: [`TlsConfig`] holds the
@@ -98,7 +98,7 @@ pub struct TlsConfig {
     /// `CertPem`/`KeyPem`, whose files are loaded as the configuration is read.
     pub identity: Option<Identity>,
     /// The Certificate Authority the server is verified against, `None` for the system CAs. Read
-    /// from `CaCert`, whose PEM file is loaded as the configuration is read.
+    /// from `CaCertPath`, whose PEM file is loaded as the configuration is read.
     pub ca_cert: Option<CertificateDer<'static>>,
     /// Override the endpoint name during SSL verification. `OverrideTargetName`.
     pub override_target_name: Option<String>,
@@ -122,7 +122,7 @@ pub(crate) struct RawTls {
     identity: RawIdentity,
     /// Path to the Certificate Authority file, in PEM format; empty for the system CAs.
     #[serde(default, deserialize_with = "crate::config_utils::text")]
-    ca_cert: String,
+    ca_cert_path: String,
     /// Accept any server certificate instead of verifying it: `1`, `true`, `yes`, `enable`,
     /// `allow` or `authorize`, and their negatives; empty for false.
     #[serde(default, deserialize_with = "crate::config_utils::text")]
@@ -211,7 +211,7 @@ impl TryFrom<RawTls> for TlsConfig {
     fn try_from(raw: RawTls) -> Result<Self, Self::Error> {
         let RawTls {
             identity,
-            ca_cert,
+            ca_cert_path,
             allow_unsafe_connection,
             override_target_name,
         } = raw;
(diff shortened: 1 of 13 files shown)


packages/rust/armonik-transport/src/tls_config.rs
```diff
@@ -111,14 +111,23 @@ pub struct TlsConfig {
...
+    schemars(transform = crate::config_utils::strip_defaults)
+)]
 #[serde(rename_all = "PascalCase")]
-struct RawTls {
+pub(crate) struct RawTls {
     #[serde(flatten)]
     identity: RawIdentity,
+    /// Path to the Certificate Authority file, in PEM format; empty for the system CAs.
     #[serde(default, deserialize_with = "crate::config_utils::text")]
     ca_cert: String,
```
wkirschenmann: This should be a path

packages/rust/armonik-transport/src/tls_config.rs
```diff
@@ -111,14 +111,23 @@ pub struct TlsConfig {
...
 #[cfg(feature = "serde")]
 #[derive(Debug, serde::Deserialize)]
+#[cfg_attr(
+    feature = "schema",
+    derive(schemars::JsonSchema),
+    schemars(transform = crate::config_utils::strip_defaults)
+)]
 #[serde(rename_all = "PascalCase")]
-struct RawTls {
+pub(crate) struct RawTls {
```
wkirschenmann, line 120: Why is RawTls required ? Can't we just use rawIdentity by providing deserialize and the tryfrom on Indentity ?

packages/rust/armonik-transport/src/tls_config.rs
wkirschenmann: This should be a named as a path

packages/rust/armonik-transport/src/tls_config.rs
```diff
@@ -131,22 +140,27 @@ struct RawTls {
 /// [`RawIdentity::load`] is what reads empty as unset, whichever shape matched.
 #[cfg(feature = "serde")]
 #[derive(Debug, serde::Deserialize)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
```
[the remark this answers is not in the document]
wkirschenmann: I'll document why you're wrong.

packages/rust/armonik-transport/src/tls_config.rs
wkirschenmann: Removable, but keeping it - and the macro is not the constraint. `embed_prefixed!` already takes the
value and schema types separately, and 3 of 5 units pass different ones, so `TlsConfig, TlsConfig`
would compile and change nothing.
`RawTls` is the document's shape, not a wrapper: all four fields differ in type from `TlsConfig`'s
(`String` against `bool`, a loaded `Option<Identity>`, a loaded `Option<CertificateDer>`,
`Option<String>`). Dropping it moves each mapping onto `TlsConfig` as a `deserialize_with` plus a
`schemars(with)` - about eight attributes on a public `#[non_exhaustive]` type, each declaring that
a field is read as something other than its declared type. `RawIdentity` stays either way.
The rule already in the code: a Raw type exists exactly when those two shapes differ. `tcp`/`http2`
have none; `tls` maps text to a bool, to a file read into DER, and to an `Option`.

